Issue #59, opened 2014-04-04
local packages not found as dependencies using python rospkg


awesomebytes:
Hello, I'm trying to revive for personal use rxbag_plugins for rxbag (as rqt_bag replacement is still not there https://github.com/ros-visualization/rqt_common_plugins/issues/85). I catkinized rxbag_plugins (https://github.com/awesomebytes/rxbag_plugins) and I pretend to use it with groovy's catkinized rx (https://github.com/ros-visualization/rx).

For finding plugins rxbag uses (https://github.com/ros-visualization/rx/blob/groovy-devel/rxbag/src/rxbag/plugins.py#L54-L55): 
```
rospack = rospkg.RosPack()
to_check = rospack.get_depends_on('rxbag', implicit=False)
```
And it seems  rospack.get_depends_on() uses parse_manifest_file() (https://github.com/ros-infrastructure/rospkg/blob/master/src/rospkg/manifest.py#L354):
```
                for d in (p.buildtool_depends + p.build_depends + p.run_depends):
                    if is_ros_package(_static_rosdep_view, d.name):
                        depends.add(d.name)
                    if is_system_dependency(_static_rosdep_view, d.name):
                        rosdeps.add(d.name)
                for name in depends:
                    manifest.depends.append(Depend(name, 'package'))
```
Which checks for the package being in the _static_rosdep_view declared before which seems to know nothing about local packages (I mean, packages in my catkin_ws/src which I can roscd into and I also can do rospack depends-on correctly).

This is probably an issue of old code and my lack of understanding of the big image running here to find dependencies. But I can do:
```
$ rospack depends-on rxbag

Output: rxbag_plugins
```
But Python:
```
import rospkg
rp = rospkg.RosPack()
manife = rp.get_manifest('rxbag_plugins')
manife.depends

Output: [catkin, std_msgs, roslib, sensor_msgs, geometry_msgs, rospy]
```

And rxbag it's in the dependences! (https://github.com/awesomebytes/rxbag_plugins/blob/master/package.xml). I tried with other packages found in my workspace that run correctly and it does not find any of them if it's not a ROS package listed in rosdep.

In the end:
1) Should I change the way rxbag searches for plugins?
2) Is this a bug in how dependencies are found?

Sorry if something is completely wrong and I'm not seeing it!

dirk-thomas:
I will close this ticket since the fix from the PR above can't be merged due to performance implications.

PR #60, opened 2014-04-05
fix package dependencies to include package names not on the RPP

dirk-thomas:
and neither released nor system dependencies. This addresses #59.

This implies that parsing a single manifest will require a full crawl of the RPP.


dirk-thomas:
Will the change in this PR is necessary to return the correct dependency classification it inflicts a significant  performance overhead and will therefore not be applied.

We will keep this PR open as a reminder of the shortcoming.

PR #66, opened 2014-08-20
Make LsbDetect check upstream LSB data as well.

gavanderhoorn:
First stab at solution for ros-infrastructure/rosdep#272.

Some slight code duplication (copy/paste from `lsb_release`), but it doesn't introduce any new dependencies.

From some googling I understand that `lsb_release` is going to be deprecated (or already is), and `platform.linux_distribution(..)` is transitioning to `/etc/os-release`. iiuc, `platform.linux_distribution(..)` is also (going to be) deprecated.

All tests pass, and my system is correctly identified as:
```
OS Name:     ubuntu
OS Version:  12.04
OS Codename: precise
```

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -130,20 +130,10 @@ class LsbDetect(OsDetector):
         raise OsNotDetected('called in incorrect OS')
 
     def load_upstream_info(self, release_file='/etc/upstream-release/lsb-release'):
-        upstream_info = read_issue(filename=release_file)
-        if upstream_info is None:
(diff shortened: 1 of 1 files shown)


tfoote:
The Fdo Detector from #88 might be a better solution. It has a reference to the parent OS in that file.

gavanderhoorn:
@tfoote: I've described some issues I encountered with using `os-release` in an earlier [comment](#issuecomment-61365606). Looks like some of those have been dealt with in #88. I'll see if I can test #88 on some derivatives.

tfoote:
Closing to look for other approaches

PR #72, opened 2014-12-23
Add support for rospack_norecurse to avoid recursing into all subdirecto...

aballier:
...ries when looking for a package.

I am installing ros into /usr, hence this searches in /usr/share. I have a lot of directories, not ROS related, into /usr/share (e.g. a full TeX Live install), and I get these timings:

Without rospack_norecurse in /usr/share:
$ time rosversion rospack
2.2.5

real    0m2.861s
user    0m1.588s
sys     0m1.267s

With rospack_norecurse in /usr/share:
$ time rosversion rospack
2.2.5

real    0m0.135s
user    0m0.100s
sys     0m0.034s

This is on a desktop with a SSD drive. On an arm board, with much less packages in /usr/share but way much slower I/O I remember hitting the 10 seconds timeout for roscore to start.

### DIFF
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -42,7 +42,7 @@ from .stack import parse_stack_file, InvalidStack
 _cache_lock = Lock()
 
 
-def list_by_path(manifest_name, path, cache):
+def list_by_path(manifest_name, path, cache, recurse=True):
     """
     List ROS stacks or packages within the specified path.
 
@@ -53,6 +53,7 @@ def list_by_path(manifest_name, path, cache):
     :param manifest_name: MANIFEST_FILE or STACK_FILE, ``str``
     :param path: path to list resources in, ``str``
     :param cache: path cache to update. Maps resource name to directory path, ``{str: str}``
+    :param recurse: search recursively in subdirectories ``bool``
     :returns: complete list of resources in ROS environment, ``[str]``
     """
     resources = []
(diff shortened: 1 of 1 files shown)


src/rospkg/rospack.py
```diff
@@ -89,9 +90,15 @@ def list_by_path(manifest_name, path, cache):
             # optimization for stacks.
             del dirs[:]
             continue #leaf     
-        elif 'rospack_nosubdirs' in files:
+        elif 'rospack_nosubdirs' in files or not recurse:
             del dirs[:]
             continue  #leaf
+        elif 'rospack_norecurse' in files:
```
wjwwood: I'm fine with the concept of `rospack_norecurse`. I don't think we have to resolve how the file gets placed in the needed directories, e.g. `/usr/share`, to agree that it could be useful just like `rospack_nosubdirs` or `CATKIN_IGNORE`.

However, for this to be merged, I agree with @dirk-thomas that it needs to be documented and other tools that currently respect `rospack_nosubdirs` also need to be updated. At least `rospack` is in that list, possibly others, though tools should use either `rospkg` (Python) or `rospack` (C++) api's rather than implement this themselves, so the list should be short.

dirk-thomas:
I will close this due to the long time of inactivity and no response to previous comments. Please feel free to command and update the patch and the PR can be reopened.

PR #82, opened 2015-07-13
Made rospkg cooperate with Funtoo Linux, a Gentoo fork.

allenh1:
The only reason this is necessary comes from a slight difference in the `/etc/gentoo-release` file. This allows the system to detect both OS's and handle them both as Gentoo (because they don't need to be handled differently).

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -370,14 +370,24 @@ class Gentoo(OsDetector):
     def get_version(self):
         if self.is_os():
             os_list = read_issue(self._release_file)
-            if os_list[0] == "Gentoo" and os_list[1] == "Base":
-                return os_list[4]
+            return os_list[4]
         raise OsNotDetected('called in incorrect OS')        
 
     def get_codename(self):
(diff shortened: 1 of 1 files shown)


tfoote:
We don't want to masquarade at this level. The os_detect should be modified to return Funtoo. Just like it returns Ubuntu and Linaro as different targets. Then in rosdep they should be aliased like this: https://github.com/ros-infrastructure/rosdep/blob/master/src/rosdep2/platforms/debian.py#L62

allenh1:
I was thinking you might say that. Just didn't want to add another OS out of the blue. :)

tfoote:
Yeah, you can probably subclass it from Gentoo and just override the one method that's different. Possibly making the name detection a function to reuse in the two methods.

allenh1:
How's that?

allenh1:
Forgot to add the definitions at the bottom for the aliasing. Added.

wjwwood:
lgtm.

tfoote:
looks great

PR #84, opened 2015-09-06
Elementary OS patch

georgepar:
Dead simple patch to make rospkg work with Elementary OS


cottsay:
The additional lines should probably be added alphabetically - all the rest of the entries are so. You should also squash these into a single commit.

georgepar:
I squashed one of the existing commits with mine so I'm closing the pull request and I will reopen a new one.

PR #86, opened 2015-09-27
added intel edison support

alessandro-aglietti:
os_detect with this patch is able to detect intel edison os when using yocto linux (default os)

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -434,15 +434,25 @@ class Windows(OsDetector):
 class IntelEdisonYoctoDetect(OsDetector):
     """
     Detect Intel Edison Yocto Linux.
-    Edison uname -a
+    
+    uname -a
     Linux myedison 3.10.17-poky-edison+ #1 SMP PREEMPT Fri Jun 19 12:06:40 CEST 2015 i686 GNU/Linux
+    
+    platform.version()
+    #1 SMP PREEMPT Fri Jun 19 12:06:40 CEST 2015
+    
+    platform.release()
+    3.10.17-poky-edison+
+    
+    platform.system()
+    Linux
     """
     def is_os(self):
-        return platform.release() == "3.10.17-poky-edison+"
+        return "edison" in platform.release()
 
     def get_version(self):
         if self.is_os():
-            return platform.version()
+            return "Yocto"
         raise OsNotDetected('called in incorrect OS')
 
     def get_codename(self):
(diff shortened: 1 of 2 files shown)


alessandro-aglietti:
I don't know because travis fail but
python -m py_compile os_detect.py
doesn't do errors

maybe a problem with the build?

src/rospkg/os_detect.py
```diff
@@ -430,6 +430,27 @@ def get_codename(self):
...
 
+
+class IntelEdisonYoctoDetect(OsDetector):
+    """
+    Detect Intel Edison Yocto Linux.
+    Edison uname -a
+    Linux myedison 3.10.17-poky-edison+ #1 SMP PREEMPT Fri Jun 19 12:06:40 CEST 2015 i686 GNU/Linux
+    """
+    def is_os(self):
+        return platform.release() == "3.10.17-poky-edison+"
```
tfoote: This detector should not be kernel version specific.

src/rospkg/os_detect.py
```diff
@@ -430,6 +430,27 @@ def get_codename(self):
...
+
+    def get_version(self):
+        if self.is_os():
+            return platform.version()
+        raise OsNotDetected('called in incorrect OS')
+
+    def get_codename(self):
+        if self.is_os():
+            return platform.release()
```
tfoote: The release string as called out above is not equivalent to a code name.

tfoote:
I believe there's a problem with python 3.2 and coverage: See @wjwwood's comment here: https://github.com/ros-infrastructure/rosdep/pull/412#issuecomment-143057022

Can you explain a little bit more about your use case and how you intend to use this. Putting things like the kernel version into the OS detection seems to be the wrong model. We don't want to require rereleasing rospkg for every new OS release.

src/rospkg/os_detect.py
alessandro-aglietti: Sorry for that, I've update the pull request

alessandro-aglietti:
Sorry for that, I've update the pull request.

The goal is support ros installation via pip on the Intel Edison Yocto Linux: https://en.wikipedia.org/wiki/Intel_Edison

tfoote:
To support this use case the best would probably be to detect a class of the linux distro(aka yocto) and not the specific hardware. Since you can also run more standard debian/ubilinux on it. It would probably be reuseable for other yocto builds.

dirk-thomas:
I will close this ticket for now since it has been stated that the detection should be based on the distro and not based on the hardware.

PR #87, opened 2015-10-02
Add OS X El Capitan to _osx_codename_map

jmtatsch:
for future compatibility with OS X El Capitan


jmtatsch:
I added a space to "el capitan" to more closely follow the naming scheme as in "mountain lion". I also opened a related pr for ros/rosdistro/pull/9596 to fix dependencies in osx-homebrew.yaml accordingly.
 @tfoote any ideas why this fails for python 3.2?

wjwwood:
Failing on Py3.2 is ok, I need to remove 3.2 and 3.3 and replace them with 3.4 anyways.

PR #88, opened 2015-10-06
Add detector to read /etc/os-release

jpgr87:
Added a detector that reads /etc/os-release and parses names, versions,
and codenames from it.  The name, FdoDetect, is based on
freedesktop.org, where the os-release spec is hosted:
http://www.freedesktop.org/software/systemd/man/os-release.html

Switched the Fedora detector from Fedora to the new FdoDetect

Signed-off-by: Rich Mattes <richmattes@gmail.com>

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -67,7 +67,7 @@ def read_issue(filename="/etc/issue"):
             return f.read().split()
     return None
 
-def read_os_release(filename = "/etc/os-release"):
+def read_os_release(filename="/etc/os-release"):
     """
     :returns: Dictonary of key value pairs from /etc/os-release, with quotes stripped from values
     """
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -149,7 +149,7 @@ class FdoDetect(OsDetector):
     """
     def __init__(self, fdo_id):
         release_info = read_os_release()
-        if release_info.has_key("ID") and release_info["ID"] == fdo_id:
+        if release_info is not None and release_info.has_key("ID") and release_info["ID"] == fdo_id:
             self.release_info = release_info
         else:
             self.release_info = None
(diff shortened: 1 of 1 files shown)


src/rospkg/os_detect.py
```diff
@@ -67,6 +67,19 @@ def read_issue(filename="/etc/issue"):
             return f.read().split()
     return None
 
+def read_os_release(filename = "/etc/os-release"):
```
wjwwood: No spaces around the `=` when defining a keyword argument: https://www.python.org/dev/peps/pep-0008/#other-recommendations

jpgr87:
Not sure why that build failed, it looks like the traceback is in the call to nosetests, not in any of the rospkg code.

src/rospkg/os_detect.py
```diff
@@ -126,6 +139,47 @@ def get_codename(self):
             return self.lsb_info[2]
         raise OsNotDetected('called in incorrect OS')
 
+class FdoDetect(OsDetector):
```
tfoote: Fdo seems too specific for this. OsRelease or something similar might make more sense. It looks like it would work on more than just fedora.

tfoote:
There's an issue with nosetest on 3.2 it's likely not your problem.

src/rospkg/os_detect.py
cottsay: Rich mentions that `Fdo` is intended to represent freedesktop.org, and isn't specific to Fedora... Since they host the spec for this file, it makes sense. What would you rather see it changed to?

jpgr87: I'm happy to change it to OsRelease or similar if that makes more sense.  I was trying to follow the LsbDetect multi-platform detector naming scheme, but it really doesn't matter to me what we call it.

tfoote: Ahh, that makes sense now. I was reading it as an oddly truncated fedora. Maybe you could put the link to the spec into the comment since that's what people in the future will see anyway.

tfoote:
Thanks for the follow up. I removed the quoted quotes. I tested this for older fedoras and it worked on 20. The Fedora class could be removed, however it's still being used by RHEL. I think this might be able to replace RHEL and CentOS detectors as well.

PR #98, opened 2016-04-14
remove dependency on catkin-sphinx

tfoote:
This makes the documentation standalone

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,7 +14,6 @@
 import sys, os
 sys.path.insert(0, os.path.abspath('../src'))
 import rospkg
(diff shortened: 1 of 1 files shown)


wjwwood:
Why? I think there was a good reason to have the theme not duplicated across many Python packages...

tfoote:
It's only used in 2 packages at the moment and we haven't released catking_sphinx since 2012. Please see https://github.com/ros-infrastructure/catkin-sphinx/issues/6

tfoote:
changed to the default theme like: https://github.com/ros-infrastructure/rosdep/pull/446

PR #99, opened 2016-05-05
Set default ROS_PACKAGE_PATH to /usr/share

scpeters:
From @jspricke's [debian patch](http://anonscm.debian.org/cgit/debian-science/packages/ros/ros-rospkg.git/tree/debian/patches/0001-Set-default-ROS_PACKAGE_PATH-to-usr-share.patch)
See also:


dirk-thomas:
Superseded by #102.

PR #100, opened 2016-05-05
Limit search in /usr/share

scpeters:
os.walk would walk all paths. Let's just use /usr/share/<pkg-name> for
ROS packages.

from @jspricke's [debian patch](http://anonscm.debian.org/cgit/debian-science/packages/ros/ros-rospkg.git/tree/debian/patches/0002-Limit-search-in-usr-share.patch)
See also:
https://github.com/ros-infrastructure/rospkg/pull/99

### DIFF
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -89,7 +89,7 @@ def list_by_path(manifest_name, path, cache):
             # optimization for stacks.
             del dirs[:]
             continue #leaf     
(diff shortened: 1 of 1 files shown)


v4hn:
:-1: This request is a (due to the additional second commit) broken version of https://github.com/ros-infrastructure/rospkg/pull/102 and should be closed as redundant

tfoote:
superseded by #102

PR #101, opened 2016-05-05
Set rosversion to Debian

scpeters:
From @jspricke's [debian patch](http://anonscm.debian.org/cgit/debian-science/packages/ros/ros-rospkg.git/tree/debian/patches/0003-Set-rosversion-to-Debian.patch)
See also:
https://github.com/ros-infrastructure/rospkg/pull/99
https://github.com/ros-infrastructure/rospkg/pull/100

### DIFF
--- a/scripts/rosversion
+++ b/scripts/rosversion
@@ -75,7 +75,7 @@ if len(sys.argv) == 2:
         else:
             distro_name = get_distro_name_from_roscore()
         if not distro_name:
-            print('<unknown>')
+            print('Debian')
         else:
             print(distro_name)
         sys.exit(0)


dirk-thomas:
`rosversion` would print `Debian` if you forget to source a setup file on every platform. I don't think it is a feasible to merge this as-is.

jspricke:
I think this is Debian only and should not go upstream.

v4hn:
:-1: Please close. 
There is no sense in setting an unknown to Debian in upstream and it adds one more item to the list of unresolved pull-requests in this repo.

PR #102, opened 2016-05-06
This set's the default ROS_PACKAGE_PATH to /usr/share

jspricke:
Initially I created these for Debian, but it should be fine to include
them upstream as well.

Replaces: #99, #100

### DIFF
--- a/src/rospkg/environment.py
+++ b/src/rospkg/environment.py
@@ -52,6 +52,9 @@ ROS_LOG_DIR      ="ROS_LOG_DIR"
 ## directory in which test result files are written
 ROS_TEST_RESULTS_DIR = "ROS_TEST_RESULTS_DIR"
 
+# FHS 4.11 package path
+ROS_SYSTEM_PACKAGE_PATH = "/usr/share"
+
 # Utilities
 def _resolve_path(p):
     """
@@ -115,9 +118,9 @@ def get_ros_package_path(env=None):
         return env.get(ROS_PACKAGE_PATH, None)
 
     if ROS_PACKAGE_PATH in env and env.get(ROS_PACKAGE_PATH):
-        return env.get(ROS_PACKAGE_PATH) + ':/usr/share'
+        return env.get(ROS_PACKAGE_PATH) + ':' + ROS_SYSTEM_PACKAGE_PATH
 
-    return '/usr/share'
+    return ROS_SYSTEM_PACKAGE_PATH
 
 def get_ros_home(env=None):
     """
--- a/src/rospkg/environment.py
+++ b/src/rospkg/environment.py
@@ -39,9 +39,10 @@ import os
 # Enviroment Variables
 
 # Global, usually set in setup
-ROS_ROOT         = "ROS_ROOT"
-ROS_PACKAGE_PATH = "ROS_PACKAGE_PATH"
-ROS_HOME         = "ROS_HOME"
+ROS_ROOT            = "ROS_ROOT"
+ROS_PACKAGE_PATH    = "ROS_PACKAGE_PATH"
+ROS_HOME            = "ROS_HOME"
+ROS_RPP_NO_DEFAULTS = "ROS_RPP_NO_DEFAULTS"
 
 # override directory path to /etc/ros
 ROS_ETC_DIR      = "ROS_ETC_DIR"
@@ -109,7 +110,14 @@ def get_ros_package_path(env=None):
     """
     if env is None:
         env = os.environ
-    return env.get(ROS_PACKAGE_PATH, '') + ':/usr/share'
+
+    if ROS_RPP_NO_DEFAULTS in env:
+        return env.get(ROS_PACKAGE_PATH, None)
+
+    if ROS_PACKAGE_PATH in env and env.get(ROS_PACKAGE_PATH):
+        return env.get(ROS_PACKAGE_PATH) + ':/usr/share'
+
+    return '/usr/share'
 
 def get_ros_home(env=None):
     """
--- a/src/rospkg/environment.py
+++ b/src/rospkg/environment.py
@@ -53,7 +53,7 @@ ROS_LOG_DIR      ="ROS_LOG_DIR"
 ROS_TEST_RESULTS_DIR = "ROS_TEST_RESULTS_DIR"
 
 # FHS 4.11 package path
-ROS_SYSTEM_PACKAGE_PATH = "/usr/share"
+ROS_SYSTEM_PACKAGE_PATH = "/usr/local/share:/usr/share"
 
 # Utilities
 def _resolve_path(p):
--- a/src/rospkg/environment.py
+++ b/src/rospkg/environment.py
@@ -39,10 +39,10 @@ import os
 # Enviroment Variables
 
 # Global, usually set in setup
-ROS_ROOT            = "ROS_ROOT"
-ROS_PACKAGE_PATH    = "ROS_PACKAGE_PATH"
-ROS_HOME            = "ROS_HOME"
-ROS_RPP_NO_DEFAULTS = "ROS_RPP_NO_DEFAULTS"
+ROS_ROOT                     = "ROS_ROOT"
+ROS_PACKAGE_PATH             = "ROS_PACKAGE_PATH"
+ROS_HOME                     = "ROS_HOME"
+ROS_PACKAGE_PATH_NO_DEFAULTS = "ROS_PACKAGE_PATH_NO_DEFAULTS"
 
 # override directory path to /etc/ros
 ROS_ETC_DIR      = "ROS_ETC_DIR"
@@ -111,7 +111,7 @@ def get_ros_package_path(env=None):
     if env is None:
         env = os.environ
 
-    if ROS_RPP_NO_DEFAULTS in env:
+    if ROS_PACKAGE_PATH_NO_DEFAULTS in env:
         return env.get(ROS_PACKAGE_PATH, None)
 
     if ROS_PACKAGE_PATH in env and env.get(ROS_PACKAGE_PATH):
--- a/src/rospkg/environment.py
+++ b/src/rospkg/environment.py
@@ -109,7 +109,7 @@ def get_ros_package_path(env=None):
     """
     if env is None:
         env = os.environ
-    return env.get(ROS_PACKAGE_PATH, None)
+    return env.get(ROS_PACKAGE_PATH, '') + ':/usr/share'
 
 def get_ros_home(env=None):
     """
--- a/test/test_rospkg_environment.py
+++ b/test/test_rospkg_environment.py
@@ -57,7 +57,7 @@ def test_get_ros_package_path():
     assert '/usr/share' == get_ros_package_path(env={})
     env = {'ROS_PACKAGE_PATH': ':'}
     assert '::/usr/share' == get_ros_package_path(env=env)
-    env = {'ROS_PACKAGE_PATH': ':', 'ROS_RPP_NO_DEFAULTS': 'True'}
+    env = {'ROS_PACKAGE_PATH': ':', 'ROS_PACKAGE_PATH_NO_DEFAULTS': 'True'}
     assert ':' == get_ros_package_path(env=env)
 
     # trip-wire tests. Cannot guarantee that ROS_PACKAGE_PATH is set
--- a/test/test_rospkg_environment.py
+++ b/test/test_rospkg_environment.py
@@ -54,13 +54,13 @@ def test_get_ros_root():
     
 def test_get_ros_package_path():
     from rospkg import get_ros_package_path
-    assert None == get_ros_package_path(env={})
-    env = {'ROS_PACKAGE_PATH': ':'}
-    assert ':' == get_ros_package_path(env=env)
+    assert ':/usr/share' == get_ros_package_path(env={})
+    env = {'ROS_PACKAGE_PATH': ''}
+    assert ':/usr/share' == get_ros_package_path(env=env)
 
     # trip-wire tests. Cannot guarantee that ROS_PACKAGE_PATH is set
     # to valid value on test machine, just make sure logic doesn't crash
-    assert os.environ.get('ROS_PACKAGE_PATH', None) == get_ros_package_path()
+    assert os.environ.get('ROS_PACKAGE_PATH', ':/usr/share') == get_ros_package_path()
 
 def test_get_log_dir():
     from rospkg import get_log_dir, get_ros_root
--- a/test/test_rospkg_environment.py
+++ b/test/test_rospkg_environment.py
@@ -54,15 +54,15 @@ def test_get_ros_root():
     
 def test_get_ros_package_path():
     from rospkg import get_ros_package_path
-    assert '/usr/share' == get_ros_package_path(env={})
+    assert '/usr/local/share:/usr/share' == get_ros_package_path(env={})
     env = {'ROS_PACKAGE_PATH': ':'}
-    assert '::/usr/share' == get_ros_package_path(env=env)
+    assert '::/usr/local/share:/usr/share' == get_ros_package_path(env=env)
     env = {'ROS_PACKAGE_PATH': ':', 'ROS_PACKAGE_PATH_NO_DEFAULTS': 'True'}
     assert ':' == get_ros_package_path(env=env)
 
     # trip-wire tests. Cannot guarantee that ROS_PACKAGE_PATH is set
     # to valid value on test machine, just make sure logic doesn't crash
-    assert os.environ.get('ROS_PACKAGE_PATH', '/usr/share') == get_ros_package_path()
+    assert os.environ.get('ROS_PACKAGE_PATH', '/usr/local/share:/usr/share') == get_ros_package_path()
 
 def test_get_log_dir():
     from rospkg import get_log_dir, get_ros_root
--- a/test/test_rospkg_environment.py
+++ b/test/test_rospkg_environment.py
@@ -54,13 +54,15 @@ def test_get_ros_root():
     
 def test_get_ros_package_path():
     from rospkg import get_ros_package_path
-    assert ':/usr/share' == get_ros_package_path(env={})
-    env = {'ROS_PACKAGE_PATH': ''}
-    assert ':/usr/share' == get_ros_package_path(env=env)
+    assert '/usr/share' == get_ros_package_path(env={})
+    env = {'ROS_PACKAGE_PATH': ':'}
+    assert '::/usr/share' == get_ros_package_path(env=env)
+    env = {'ROS_PACKAGE_PATH': ':', 'ROS_RPP_NO_DEFAULTS': 'True'}
+    assert ':' == get_ros_package_path(env=env)
 
     # trip-wire tests. Cannot guarantee that ROS_PACKAGE_PATH is set
     # to valid value on test machine, just make sure logic doesn't crash
-    assert os.environ.get('ROS_PACKAGE_PATH', ':/usr/share') == get_ros_package_path()
+    assert os.environ.get('ROS_PACKAGE_PATH', '/usr/share') == get_ros_package_path()
 
 def test_get_log_dir():
     from rospkg import get_log_dir, get_ros_root
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -89,7 +89,7 @@ def list_by_path(manifest_name, path, cache):
             # optimization for stacks.
             del dirs[:]
             continue #leaf     
-        elif 'rospack_nosubdirs' in files:
+        elif 'rospack_nosubdirs' in files or d.startswith('/usr/share/'):
             del dirs[:]
             continue  #leaf
         # remove hidden dirs (esp. .svn/.git)
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -35,7 +35,7 @@ from threading import Lock
 from xml.etree.cElementTree import ElementTree
 
 from .common import MANIFEST_FILE, PACKAGE_FILE, STACK_FILE, ResourceNotFound
-from .environment import get_ros_paths
+from .environment import get_ros_paths, ROS_SYSTEM_PACKAGE_PATH
 from .manifest import parse_manifest_file, InvalidManifest
 from .stack import parse_stack_file, InvalidStack
 
@@ -89,7 +89,7 @@ def list_by_path(manifest_name, path, cache):
             # optimization for stacks.
             del dirs[:]
             continue #leaf     
-        elif 'rospack_nosubdirs' in files or d.startswith('/usr/share/'):
+        elif 'rospack_nosubdirs' in files or d.startswith(ROS_SYSTEM_PACKAGE_PATH):
             del dirs[:]
             continue  #leaf
         # remove hidden dirs (esp. .svn/.git)


v4hn:
The tests pass and this request adds support to install ros-packages to /usr
in an easy to understand way.

:+1: please merge

dirk-thomas:
Imo merging this patch as-is has the following problems:

* it only addresses Python, `rospack` wouldn't behave the same way and therefore be inconsistent
* appending `/usr/share` unconditionally might lead to unexpected behavior if you explicitly set the RPP to something excluding `/usr/share` but it still considers that location.

v4hn:
> appending /usr/share unconditionally might lead to unexpected behavior if you explicitly set the RPP to something excluding /usr/share but it still considers that location.

I don't think this is a problem. If you believe this is a block for this patch, how would you propose to change it?

dirk-thomas:
> > appending /usr/share unconditionally might lead to unexpected behavior if you explicitly set the RPP to something excluding /usr/share but it still considers that location.

> I don't think this is a problem. If you believe this is a block for this patch, how would you propose to change it?

It potentially finds packages in `/usr/local` which it shouldn't - sounds like a problem to me. Sorry, I don't have any proposal how to satisfy the Debian use case.

jspricke:
* Dirk Thomas <notifications@github.com> [2016-08-10 14:23]:
> It potentially finds packages in `/usr/local` which it shouldn't - sounds like a problem to me. Sorry, I don't have any proposal how to satisfy the Debian use case.

I'm sorry, but how should that happen if I add `/usr/share`?

dirk-thomas:
I just misspelled it - `usr/local` was supposed to be `/usr/share`.

src/rospkg/environment.py
```diff
@@ -109,7 +109,7 @@ def get_ros_package_path(env=None):
     """
     if env is None:
         env = os.environ
-    return env.get(ROS_PACKAGE_PATH, None)
+    return env.get(ROS_PACKAGE_PATH, '') + ':/usr/share'
```
wjwwood: You could consider doing this like:

```python
return env.get(ROS_PACKAGE_PATH, ':/usr/share')
```

The upside is that you could use rospkg without considering `/usr/share` by manually setting the `ROS_PACKAGE_PATH`. The downside is that people who set the `ROS_PACKAGE_PATH` need to make sure to add `/usr/share` to it when they're using packages from there.

I think it makes more sense for the Debian packages to set the `ROS_PACKAGE_PATH` to by inserting it into some shell rc file or something. That way there is visibility into what is being searched. Also, it would prevent the inconsistency between `rospkg` (Python) and `rospack` (C++).

test/test_rospkg_environment.py
```diff
@@ -54,13 +54,13 @@ def test_get_ros_root():
     
 def test_get_ros_package_path():
     from rospkg import get_ros_package_path
-    assert None == get_ros_package_path(env={})
-    env = {'ROS_PACKAGE_PATH': ':'}
-    assert ':' == get_ros_package_path(env=env)
+    assert ':/usr/share' == get_ros_package_path(env={})
+    env = {'ROS_PACKAGE_PATH': ''}
```
wjwwood: I don't think removing the `:` in the example environment is the right way to make this test pass. In the current patch, if `ROS_PACKAGE_PATH` is set to `:` then `get_ros_package_path()` would return `::/usr/share`. The test should reflect that or the code should ensure that no redundant `:` are left in the result.

src/rospkg/environment.py
wjwwood: You could also just generate the normal catkin `setup.*sh` files and install them to something like `/usr/share/ros/setup.*sh`. Then to use ros you need to source this file, but that's not a huge burden, imo. Plenty of other software have a similar mechanism to setup the environment before being used. You could still source this file by default by using `/etc/profile.d` or similar.

jspricke: That's against [Debian Policy §9.9](http://www.debian.org/doc/debian-policy/ch-opersys.html#s9.9)

wjwwood: That's kind of a idealistic policy, imo. It seems to put the goals of the Debian project ahead of the software they're integrating, i.e. that software should work without any configuration or changes to the environment. By comparison, Red Hat has `/etc/profile.d` and they manage to struggle along (https://bugs.launchpad.net/ubuntu/+source/base-files/+bug/18808). Most of the time that's fine, but in this case it makes the tool (`rospkg`) unable to function outside of `/usr`, which kind of sucks. The policy's proposed work around of replacing all commands with a shell wrapper script would, I think, introduce a noticeable delay for commands used in tab-completion. And what about software that provides shell functions or tab completion helpers? They don't work for all shells (which was one of the justifications of the policy), how are they different from setting an env variable (which is extremely portable btw)?

Given that the policy is not going to change, then I'd still opt for having the normal setup files and let users source them and/or add them to their shell rc files as they like. This behavior (of always having `/usr/share` in the RPP) is unfortunately a deal breaker for using the debian version of `rospkg` with ROS installed to `/opt`.

jspricke: Regarding rospack, I have patches as well (you can find them [here](https://anonscm.debian.org/cgit/debian-science/packages/ros/ros-rospack.git/tree/debian/patches)) and I will submit them once we found a solution over here, as they are quite similar.

jspricke: * William Woodall <notifications@github.com> [2016-08-10 15:02]:
> Given that the policy is not going to change, then I'd still opt for having the normal setup files and let users source them and/or add them to their shell rc files as they like. This behavior (of always having `/usr/share` in the RPP) is unfortunately a deal breaker for using the debian version of `rospkg` with ROS installed to `/opt`.

Can you give an example where this could break so we can find a fix for
that?

wjwwood: The only other compromise that I could think of is that this patch could have some mechanism for not including `/usr/share` in the RPP (by an environment variable maybe) and we could trigger that mechanism when using ROS from `/opt`.

wjwwood: > Can you give an example where this could break so we can find a fix for that?

Sure, I think I can demonstrate how it can be logically inconsistent, and a come up with a hypothetical scenario where that inconsistency could actually actually break something.

If you have packages A, B, and C installed from Ubuntu (into `/usr`) and you have A, B, and D installed in `/opt/ros/kinetic`, then you do `rospack list` you'll get A, B, and C. However, if you source the `/opt/ros/kinetic/setup.bash` and do `rospack list` you'll get A, B, C, and D. Where C comes from `/usr` and the others point to `/opt/ros/kinetic`. Let's say that both C and D build against (use headers from and link against) B.

This can be a problem if, for example, you build a package locally called E which uses both C and D.

First consider you do this in a clean shell. The build should fail because D is not found.

Then consider you try to build it after sourcing the `/opt/ros/kinetic/setup.bash` file. Then the build does not fail because it finds C in `/usr` and D in `/opt/ros/kinetic`. The "expected" behavior would be that the build fails because C is not found, but that's not what happens. However, even if the build works (might not due to differences in headers between versions of B), it very likely will have problems when running since C is built against B in `/usr` and D is built against B in `/opt/ros/kinetic` (let's call that B'). If B and B' are not ABI compatible, then you've got a problem.

This same issue can be extended to Python code, cmake modules, locating resources from other packages, etc.

---

There is also another issue that comes up that's sort of the reverse of the above, where you get the `rospkg` from OSRF rather than Ubuntu because it's newer (for example). That leads to failures to run things like `roslaunch` when you have ROS installed from both places. I think I sent you emails on this issue, but running `roslaunch` I got:

```
william@ubuntu:~$ roscore
WARNING: unable to configure logging. No log files will be generated
Checking log directory for disk usage. This may take awhile.
Press Ctrl-C to interrupt
Done checking log file disk usage. Usage is <1GB.
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/roslaunch/__init__.py", line 307, in main
    p.start()
  File "/usr/lib/python2.7/dist-packages/roslaunch/parent.py", line 268, in start
    self._start_infrastructure()
  File "/usr/lib/python2.7/dist-packages/roslaunch/parent.py", line 217, in _start_infrastructure
    self._load_config()
  File "/usr/lib/python2.7/dist-packages/roslaunch/parent.py", line 132, in _load_config
    roslaunch_strs=self.roslaunch_strs, verbose=self.verbose)
  File "/usr/lib/python2.7/dist-packages/roslaunch/config.py", line 445, in load_config_default
    load_roscore(loader, config, verbose=verbose)
  File "/usr/lib/python2.7/dist-packages/roslaunch/config.py", line 92, in load_roscore
    f_roscore = get_roscore_filename()
  File "/usr/lib/python2.7/dist-packages/roslaunch/config.py", line 84, in get_roscore_filename
    return os.path.join(r.get_path('roslaunch'), 'resources', 'roscore.xml')
  File "/usr/lib/python2.7/dist-packages/rospkg/rospack.py", line 200, in get_path
    raise ResourceNotFound(name, ros_paths=self._ros_paths)
ResourceNotFound: roslaunch
```

This happened because the version of `rospkg` that got installed didn't have this patch and so it couldn't find `roslaunch` in `/usr`. That's why I was so interested in finding a way to satisfy the Debian use case with a patch that we could merge upstream -- to avoid this issue.

---

For both of these reasons we've resorted to recommending people not install both at the same time. In practice this is probably fine, but not an ideal outcome.

jspricke: * William Woodall <notifications@github.com> [2016-08-10 15:35]:
> Sure, I think I can demonstrate how it can be logically inconsistent, and a come up with a hypothetical scenario where that inconsistency could actually actually break something.
> 
> If you have packages A, B, and C installed from Ubuntu (into `/usr`) and you have A, B, and D installed in `/opt/ros/kinetic`, then you do `rospack list` you'll get A, B, and C. However, if you source the `/opt/ros/kinetic/setup.bash` and do `rospack list` you'll get A, B, C, and D. Where C comes from `/usr` and the others point to `/opt/ros/kinetic`. Let's say that both C and D build against (use headers from and link against) B.
> 
> This can be a problem if, for example, you build a package locally called E which uses both C and D.
> 
> First consider you do this in a clean shell. The build should fail because D is not found.
> 
> Then consider you try to build it after sourcing the `/opt/ros/kinetic/setup.bash` file. Then the build does not fail because it finds C in `/usr` and D in `/opt/ros/kinetic`. The "expected" behavior would be that the build fails because C is not found, but that's not what happens. However, even if the build works (might not due to differences in headers between versions of B), it very likely will have problems when running since C is built against B in `/usr` and D is built against B in `/opt/ros/kinetic` (let's call that B'). If B and B' are not ABI compatible, then you've got a problem.
> 
> This same issue can be extended to Python code, cmake modules, locating resources from other packages, etc.

That's not a ROS problem (or one with this patch after all). Just
replace "package" with "plain C library" and you get the same problem.
Generally it is a problem if users try to mix different package sources.

I don't think we can find a solution here, as long as people are able to
install conflicting versions at the same time (I've worked on a more
general solution to this already, but I don't think this is the right
place to discuss it).

> This happened because the version of `rospkg` that got installed didn't have this patch and so it couldn't find `roslaunch` in `/usr`. That's why I was so interested in finding a way to satisfy the Debian use case with a patch that we could merge upstream -- to avoid this issue.

Yes, so we agree that it would be great to merge this patch upstream and
don't run into it again :).

wjwwood: Just to summarize again the options as I see them:

- Implicitly insert `/usr/local` unless some heuristic is met, e.g. don't insert it if:
  - `rospack` can be found outside of `/usr` or
  - `ROS_NOT_DEBIAN` is set in the environment (this could be set in `/opt/ros/kinetic/setup.*sh`)
- Don't change `rospkg`, but just install the normal setup files in an appropriate location, have users source them before using (violates 9.9)
- Have a different name for the packages.ros.org version of `rospkg` (not sure this is a tractable solution)

jspricke: * William Woodall <notifications@github.com> [2016-08-10 15:53]:
>   - `ROS_NOT_DEBIAN` is set in the environment (this could be set in `/opt/ros/kinetic/setup.*sh`)

I would propose to name it `RPP_NO_DEFAULTS`, as this is not about
Debian ;). Also I see this similar to `gcc -nostdinc` and alike. If you
agree, I will implement this tomorrow.

wjwwood: Yeah, `RPP_NO_DEFAULTS ` is more generic, +1 for the name.

It's still pretty intrusive in our system in that we need to modify some package to set this env variable when being sourced (I was thinking in `roslib`) and then you need to disable that when you're packaging the same package for Debian. For newer ROS distributions that should be ok, but it is not going to be possible to introduce it in EOL ROS distros, e.g. Groovy or Hydro, but that should be fine since this doesn't really show up until Xenial, i.e. there shouldn't be ROS packages in `/usr` before that so adding it to the RPP by default will at worst slowdown searches for package manifests.

I think that sounds like the best compromise that I've heard. @dirk-thomas what do you think about having `/usr` in by default and having a way to disable those defaults by env variable?

dirk-thomas: I don't like the idea that we need to set an additional environment variable for this reason.

If we would do so none of the current ROS users has that environment variable set. We can push releases which update the setup files under `/opt/ros/distro` to always set that variable. But what do we do about distros which we are not releasing binary packages for anymore? Usually we try hard to maintain the ability to continue building them from source. This would imply changing behavior on those (hopefully it doesn't break anything - but I would try to avoid that).

The `RPP_NO_DEFAULTS` only addresses the case of a single environment variable. But ROS has many more and any package can contribute more using environment hooks. How are those being set on Debian? I personally think sourcing a file on Debian is the better approach.

wjwwood: I agree that sourcing the setup files is the better approach, but given that Debian policy 9.9 I don't think that's acceptable.

I don't think it's a problem for older distros. Take Hydro, for example, Hydro users will get the updated `rospkg` that inserts `/usr/share` implicitly, but not the binary packages that set `RPP_NO_DEFAULTS`. But that shouldn't be a problem since there are no packages in `/usr` to mess with Hydro users. I guess there's a chance for breaking things in a way that I can't think of right now and for slower performance when searching for packages, but that's all. There's no obvious issue that I can see.

For packages with other env variables, they will need to be patched as they are put into Debian one-by-one. It doesn't affect our packages, but it does mean more work for the Debian packagers.

dirk-thomas: If all the other environment variable (e.g. `ROS_ROOT`, `ROS_MASTER_URI`, `ROS_DISTRO`, the completion scripts, and every custom env var) need to be handled explicitly (implying patches on the Debian side) why should the RPP be any different? Is it really worth avoiding a simple patch which appends the prefix as proposed in this PR for `rospkg` as well as `rospack`?

tfoote: The other option which allows things to work by default but won't interfere with the normal flow is that the environment variables only have a default value if unset. So in the pure `/usr` case everything works with the default because there's no environment variable set. But if you set RPP it uses the RPP. And if you want to overlay on top of `/usr` you need to add it to your RPP. This means that it only will effect people who are advanced enough to want to create an overlay/modify the RPP. For those users the standard setup file could be provided in `/usr/share` as a way to load the defaults to be able to extend them. Such that to extend the default environment you can source the explicit default environment and then extend it.

wjwwood: > Is it really worth avoiding a simple patch which appends the prefix as proposed in this PR for rospkg as well as rospack?

The difference is that `rospkg` is installed out side of `/opt` and conflicts with the one in Ubuntu. You install one or the other, not both. I mentioned a specific case where this was a problem above. This is one of the only, if not the only, patch that I think makes a big difference in terms of usability when both are installed.

> The other option which allows things to work by default but won't interfere with the normal flow is that the environment variables only have a default value if unset.

You mean this? https://github.com/ros-infrastructure/rospkg/pull/102#discussion_r74334509

I don't think that's a good user experience. Consider this common practice of ROS users:

```
$ rospack list
... lots of things
$ export ROS_PACKAGE_PATH=$ROS_PACKAGE_PATH:`pwd`
$ rospack list
... only the packages under the current directory (not the packages in `/usr`)
```

(assume I'm invoking the same interface in `rospkg` rather than using `rospack list`)

jspricke:
I've added a patch to respect the environment variable. Node that I added a `ROS_` prefix to it to make it consistent with the rest and put it into the ROS namespace.

src/rospkg/environment.py
wjwwood: Bump, @tfoote and @dirk-thomas I responded to both of your points, can you comment if you agree or disagree?

dirk-thomas: While only one version of `rospkg` can be installed that is not the case with `rospack`. Since there behavior must be the same that means that we also have to modify `ros-<distro>-rospack` despite the fact that it is independent of Debian.

Also the new env var needs to be set by some `ros-<distro>-?` package. While that package can depend on a new version number of `rospkg` there is no way to prevent users to update only `rospkg` but not that ros package. The result will be that on Ubuntu the user now has `/usr/share` in his RPP which will significantly increase the crawling time. This will also be the default case for distributions we have are not building anymore since `rospkg` will be deployed across all ROS distros.

Imo there are too many cons involved with this change. Therefore I still don't think that ROS should introduce yet another environment variable just to toggle this behavior necessary for Debian. But if the majority wants to go ahead with this I won't stand in the way.

src/rospkg/environment.py
```diff
@@ -39,9 +39,10 @@
 # Enviroment Variables
 
 # Global, usually set in setup
-ROS_ROOT         = "ROS_ROOT"
-ROS_PACKAGE_PATH = "ROS_PACKAGE_PATH"
-ROS_HOME         = "ROS_HOME"
+ROS_ROOT            = "ROS_ROOT"
+ROS_PACKAGE_PATH    = "ROS_PACKAGE_PATH"
+ROS_HOME            = "ROS_HOME"
+ROS_RPP_NO_DEFAULTS = "ROS_RPP_NO_DEFAULTS"
```
dirk-thomas: The abbreviation `RPP` is not used anywhere. Therefore I would suggest sticking to a more verbose name: `ROS_PACKAGE_PATH_NO_DEFAULTS`.

src/rospkg/environment.py
tfoote: I agree that it's not a great user experience, but there must be a way to keep `/usr` off the RPP. Our tools all use the RPP as a search path. If it's unset the tools are all designed to return nothing since there is no path to search. In the case that RPP is set, it should always use exactly what's set. 

The debian guidelines say that it cannot need an environment so my proposed compromise is that an unset RPP defaults to use `/usr` if unset. The one case you called out is a little odd, but if you echoed RPP

```
$ echo $ROS_PACKAGE_PATH
$ export ROS_PACKAGE_PATH=$ROS_PACKAGE_PATH:`pwd`
$ echo $ROS_PACKAGE_PATH
:/path/to/pwd
```

It's very clear what your setting. There's no way for us to provide introspection to the default value because it's debian's requirement that it magically work without setting the value.

I would advocate for using the `RPP_NO_DEFAULTS` and that would turn off the "default" which is only used when the RPP is set. If we require setting `RPP_NO_DEFAULTS` even if you've explicitly set the RPP in your setup script we will break every single existing environment setup file and example setup file because suddenly their environment will now have unknown packages from `/usr` unless they set `RPP_NO_DEFAULTS`. Whereas if the default is only when RPP is unset, then the only people affected by the default value are the people not setting their RPP. This is the use case that the debian guidelines are targeted at. 

We just have to document that you cannot naively extend an unset path.

wjwwood: > Since there behavior must be the same

Must it? In the proposal, `rospkg` will act like the debian version of `rospack` when the env variable is not set and like the packages.ros.org version of `rospack` when it is. I see no reason to have `rospack` look for the env variable and change behavior like `rospkg`.

> While that package can depend on a new version number of rospkg there is no way to prevent users to update only rospkg but not that ros package. The result will be that on Ubuntu the user now has /usr/share in his RPP which will significantly increase the crawling time. This will also be the default case for distributions we have are not building anymore since rospkg will be deployed across all ROS distros.

I agree with that assessment, but the alternative is to leave this know issue unresolved. The workaround for `/usr/share` being on the path is to manually set the env variable (for older distros that won't be updated) and to update the necessary ROS package for current distros. By comparison, the workaround for the other bug is to uninstall all of the packages.ros.org packages (including the rospkg), removing packages.ros.org from apt-get, then reinstalling rospkg. So, from my perspective this proposed action is a little messy, but also addresses the issue that is more difficult for users to deal with.

> I agree that it's not a great user experience, but there must be a way to keep /usr off the RPP.

In the current proposal there is. If you set `ROS_RPP_NO_DEFAULTS` in the environment then `/usr/share` will not be on the package path.

> It's very clear what your setting. There's no way for us to provide introspection to the default value because it's debian's requirement that it magically work without setting the value.

I agree with you, but we cannot get unsurprising behavior without setting an environment variable, but that is not allowed under 9.9 as @jspricke pointed out. So I think that between `/usr/share` always being implicitly on the RPP versus it being on the RPP until you set it, I'd prefer the former.

> I would advocate for using the RPP_NO_DEFAULTS and that would turn off the "default" which is only used when the RPP is set.

If that's the case, then you never need to set the RPP_NO_DEFAULTS, right? Because the `/opt/ros/kinetic` setup files will always set the RPP.

> If we require setting RPP_NO_DEFAULTS even if you've explicitly set the RPP in your setup script we will break every single existing environment setup file and example setup file because suddenly their environment will now have unknown packages from /usr unless they set RPP_NO_DEFAULTS.

Which setup files will have this problem? As long as they have sourced the `/opt/ros/kinetic/setup.*sh` file and they have `ros-core` installed, then `ROS_RPP_NO_DEFAULTS` will be set. There may be some cases, I just cannot think of them now.

I understand where you're coming from, but I still have a slight preference for `/usr/share` always being on the RPP, despite the fact that my first suggestion was to only add it if the RPP was unset.

src/rospkg/environment.py
wjwwood: +1

src/rospkg/environment.py
jspricke: > Must it? In the proposal, `rospkg` will act like the debian version of `rospack` when the env variable is not set and like the packages.ros.org version of `rospack` when it is. I see no reason to have `rospack` look for the env variable and change behavior like `rospkg`.

I have the same patch for rospack waiting over here...

> While that package can depend on a new version number of rospkg there is no way to prevent users to update only rospkg but not that ros package. The result will be that on Ubuntu the user now has /usr/share in his RPP which will significantly increase the crawling time. This will also be the default case for distributions we have are not building anymore since rospkg will be deployed across all ROS distros.

Actually the second commit limits the crawl, so this is actually almost for
free.

I've renamed the variable as proposed.

wjwwood: > I have the same patch for rospack waiting over here...

Regardless, it isn't required since the two versions of `rospack` can be installed side-by-side without consequence, but that is not true for `rospkg`.

tfoote: So looking at the [debian policy 9.9](https://www.debian.org/doc/debian-policy/ch-opersys.html#s9.9) closely it's actually pretty clear about how to set things. It says: 
`If a program usually depends on environment variables for its configuration, the program should be changed to fall back to a reasonable default configuration if these environment variables are not present.`

And it gives an example: 
```
     #!/bin/sh
     BAR=${BAR:-/var/lib/fubar}
     export BAR
     exec /usr/lib/foo/foo "$@"
```
Which uses the syntax `:-` which only sets the variable if it's unset. [description here](https://unix.stackexchange.com/questions/122845/using-a-b-for-variable-assignment-in-scripts)

The sentence does not say "if and only if", but the example explicitly uses that logic.

wjwwood: That's true. I hadn't noticed that before. If that's what we want to do then we don't need the `ROS_PACKAGE_PATH_NO_DEFAULTS` option, because the `ROS_PACKAGE_PATH` will always be set by the `/opt/ros/kinetic/setup.*sh` files, and so when they are sourced the `/usr/share` will not be added.

All that being said, I still think behaving like `gcc` is less surprising, i.e. some include paths are always added unless you tell it not to include them automatically and adding your own include paths do not remove those default ones.

v4hn:
> Which uses the syntax :- which only sets the variable if it's unset. description here
The sentence does not say "if and only if", but the example explicitly uses that logic.

That example doesn't consider *a list* of paths, just one path in a variable.
To adhere to the principle of least surprise, default paths should not be deleted if any other path is added to the list. The policy is meant to help enforce that principle: Programs should "just work" after boot without adding custom paths to environment variables (and they should still work if you add one more path to a list).

As it stands, this request adds `/usr/share` to the ROS_PACKAGE_PATH with minimal overhead because `/usr/share` [is not walked recursively](https://github.com/ros-infrastructure/rospkg/pull/102/commits/0a5088aaf6509812fb1588c651a6c3c014bfa4c6). So as long as nobody installs ROS modules to `/usr` nothing changes really. *If* you install modules to `/usr` these are being picked up by everything else. This is what the typical Linux-user would expect, just as `boost`, `poco`, `assimp`, `eigen3` and everything else is picked up if they are installed to `/usr`.

To nevertheless allow to exclude the default path the way cmake allows to exclude the default paths via `NO_DEFAULT_PATH`, @jspricke added the `ROS_PACKAGE_PATH_NO_DEFAULTS` flag.
In my very humble opinion this should never ever be done and implies poor dependency design.
Nevertheless it's perfectly fine to allow for the possiblility.

In my opinion this request can be merged as-is.

With the osrf packages in `/opt` I do not see a problem when the magic "no-defaults" variable is not set. As OSRF packages *all* ROS-packages these always override their respective counterparts in `/usr`. Partial overlays are *always* problematic and if someone installs a package to `/usr` that is not available in the overlaying workspaces and then mixes modules from both workspaces as dependencies, it is usually perceived to be the user's fault if this breaks.
If you would like to enforce the `ROS_PACKAGE_PATH_NO_DEFAULTS` variable in osrf's ROS packages, could you please give a concrete example of why this is needed?

Also, I'm a bit tired of hearing "we are not Debian" phrased one way or another. This makes me feel really sad for two reasons. (1) This should really be "we are not/don't want to be a POSIX/Unix conform Open Source Project". Debian just tries to enforce standards and quasi-standards in the Unix/FOSS domain rigorously. (2) POSIX/Unix conform installations are/should **not** be a special case of the ROS framework. Instead, ROS should be a special case of a Linux library/framework.

wjwwood:
> Partial overlays are always problematic and if someone installs a package to /usr that is not available in the overlaying workspaces and then mixes modules from both workspaces as dependencies, it is usually perceived to be the user's fault if this breaks.

That's true, and I agree with @jspricke's analysis that overlay ROS packages is not very different from overlaying shared libraries. I also agree with you that it's a problem that people run into from time to time, and when they do it's very difficult to debug. So if we can avoid it easily then I think we should. If I could easily avoid all partial overlay problems as easily I would do that too. I believe that setting the `ROS_PACKAGE_PATH_NO_DEFAULTS` in one of our core ROS packages will be a better experience for our users out-of-the-box.

> If you would like to enforce the ROS_PACKAGE_PATH_NO_DEFAULTS variable in osrf's ROS packages, could you please give a concrete example of why this is needed?

I have a package that is building a URDF of a robot using xacro and it does so by including a `urdf.xacro` file from a `robot_description` like package. If I have installed version A of this description package to `/usr`, but have not installed it to `/opt` (and I'm intending to using `/opt` to build my package). I expect version B of the description package to be found because that's what is released into the ROS distro I have installed in `/opt`. However, when I include this `urdf.xacro` file in my xacro file, it finds the file, but it's the wrong version. If I'm lucky the arguments to the xacro file have changed and I get an error. If I am unlucky, then it just works, but the content of the description is wrong.

Now you can argue that is just what happens and the user needs deal with this scenario, but I'd like to avoid these type of problems when possible. I think it's a much better result in the example above if the description package is not found at all, despite the fact that a version of it can be found in `/usr`. Then I discover my error directly and can fix it. I'm not saying this will happen often, but it's also not a considerable burden for us to set one more environment variable either.

v4hn:
Ok, I believe your example is invalid, but I just deleted 12 lines of explanation because this is not too relevant for now.

In my opinion this request is ready to be merged.
What exactly has to be done to get this merged? And who can do it?

wjwwood:
I'd like to have some acknowledgement of consensus from the other maintainers before taking action. Then someone, perhaps me, should open the corresponding pull request to set the env variable in one of our packages. Like roslib? Those ROS packages need to be released and public before releasing rospkg with this change.

dirk-thomas:
My general opinion hasn't changed that I am not really in favor of adding any env var for this use case.

But if you want to move forward with an additional environment variable I would like to point out that the current patch embeds the string `/usr/share` six times. The patch should aim to encode this constant in as few places as possible. And if more than one location is needed define a constant for it.

I also want to suggest an alternative to the "boolean" env var `ROS_PACKAGE_PATH_NO_DEFAULTS`. Instead we could define an env variable which actually contains the always appended suffix. If the variable is not defined its default value is `/usr/share` which achieves the goal on Debian. `roslib` (e.g.) would export it with an empty value which keeps the RPP as it is. That makes the appended value explicit and also allows to provide any value in case someone would like a different location. Since that proposed env var doesn't contain a default value (since it is always appended to the RPP) it could be named `ROS_PACKAGE_PATH_SUFFIX` (or something similar).

tfoote:
I agree that the boolean to turn off the behavior as an environment variable is awkward, and also ends up with a hardcoded path in the code that cannot be adapted for other use cases. If we want to do that how about something like `ROS_PACKAGE_PATH_PERSISTENT` And if we're doing this we should be making helper functions that generate the resultant `ROS_PACKAGE_PATH` and we will need to modify anywhere that is getting the `ROS_PACKAGE_PATH` to use the helper functions that do the merging. 

Also another place that this will notably effect is the logic in the catkin generated setup files which generate the `ROS_PACKAGE_PATH` programmatically and do not just append/prepend.

Overall I feel that this is making the system more complicated having to manage two environment variables in parallel. We have one environment variable of variable length, and simply setting a default value if unset seems to me to be much easier to explain to our users than, "here's a `_PATH` variable just like you'd expect, and it gets merged with this other environment variable, to be named, following these policies: ..." 

Whereas if you simply say "The `ROS_PACKAGE_PATH` defines the search path, if it's unset it defaults to `/usr/share`" It's very clear and simple. As @wjwwood it does break one use case which is manually extending the RPP using our standard shorthand: 
```export ROS_PACKAGE_PATH=$ROS_PACKAGE_PATH:`pwd`
```
But that is easy to understand as you can easily see that previously the RPP was unset and now it's set to `pwd` and by the above policy it will only search `pwd`. 

If you already have sourced a setup file or have updated your RPP this will not be a problem. I already have a bash alias to extend my RPP quicker than the above we could simply change our recommended the oneliner to cover that case by adding the default value conditionally. 
```
alias rosaddpwd='export ROS_PACKAGE_PATH=`pwd`:${ROS_PACKAGE_PATH:-/usr/share}'
```

We could even distribute that as a bash/sh function to standardize it instead of having everyone have a similar alias.

tfoote:
Talking with @dirk-thomas we also need a way to suppress the recursive search and we should do that in a non-hard coded way, but with a default to meet 9.9 we will need to add something like `ROSPACK_NORECURSION_PATH` with a default of `/usr/share` 

For reference each directory walking on my machine takes 5-8us empty, but 10-20us with files in the directory. Making a walk take an order of magnitude longer than walking my 348 kinetic packages in /opt. I have a little over 400 subdirectories of `/usr/share` but recursively there's over 15000 directories which slows rospack profile way down into taking longer than a second. 

My test results are at: https://gist.github.com/tfoote/dce2eae42b3c8657154e8893cdc105a0 and note that this is on a modern i7 with SSD.

jspricke:
* Dirk Thomas <notifications@github.com> [2016-08-17 14:15]:
> But if you want to move forward with an additional environment variable I would like to point out that the current patch embeds the string `/usr/share` six times. The patch should aim to encode this constant in as few places as possible. And if more than one location is needed define a constant for it.

Changed. Note that half of them are part of unit test, where I think
hard coding the result is the right way.

> I also want to suggest an alternative to the "boolean" env var `ROS_PACKAGE_PATH_NO_DEFAULTS`. Instead we could define an env variable which actually contains the always appended suffix. If the variable is not defined its default value is `/usr/share` which achieves the goal on Debian. `roslib` (e.g.) would export it with an empty value which keeps the RPP as it is. That makes the appended value explicit and also allows to provide any value in case someone would like a different location. Since that proposed env var doesn't contain a default value (since it is always appended to the RPP) it could be named `ROS_PACKAGE_PATH_SUFFIX` (or something similar).

As Tully noted, I don't see any use case of this that wouldn't be easier
handled with the current proposal. Also note that an empty environment
variable is rather fragile as not all shells differentiate it from an
undefined variable (Python does, though).

* Tully Foote <notifications@github.com> [2016-08-17 17:23]:
> Talking with @dirk-thomas we also need a way to suppress the recursive search and we should do that in a non-hard coded way, but with a default to meet 9.9 we will need to add something like `ROSPACK_NORECURSION_PATH` with a default of `/usr/share` 

Feel free to send an other pull request, once this is accepted ;).

Cheers Jochen

v4hn:
* Tully Foote wrote:

> Overall I feel that this [new alternative] is making the system more complicated having to manage two environment variables in parallel.

On the contrary, the current proposal *is* very simple (effectively 12 lines of code including tests), is tested, works, and does not introduce the weird quirk that the RPP might change if you append to it.
The additional boolean variable can be ignored by all ROS users and is only required because the maintainers want a way to override the standard-conform behavior (which we conceded as a valid point for the moment).
OSRF has the option to just remove the variable at some point in time if they want to build on top of ROS packages installed in /usr, just as you build on top of a system-wide Python installation, Qt, etc.

Leaving aside one line of code that has to be added to some osrf-ROS module to set the boolean flag, this request is a code patch ready to merge and does not rely on the English language - in contrast to suggestions by you and dirk.

>  and also ends up with a hardcoded path in the code that cannot be adapted for other use cases.

To reiterate: POSIX conform installation *should not be* a special case of a catkin installation.
The File System Hierarchy is *not* some random path. It's a very old standard ([in the current version published in 2004](http://www.pathname.com/fhs/) and catkin uses it for system versions of cmake, Python, Qt, etc. You *should not* adapt the required code "for other use cases".

tfoote:
>> Tully Foote wrote:
>> Overall I feel that this [new alternative] is making the system more complicated having to manage two environment variables in parallel.

> On the contrary, the current proposal is very simple (effectively 12 lines of code including tests), is tested, works, and does not introduce the weird quirk that the RPP might change if you append to it.
> The additional boolean variable can be ignored by all ROS users and is only required because the maintainers want a way to override the standard-conform behavior (which we conceded as a valid point for the moment).
> OSRF has the option to just remove the variable at some point in time if they want to build on top of ROS packages installed in /usr, just as you build on top of a system-wide Python installation, Qt, etc.
> 
> Leaving aside one line of code that has to be added to some osrf-ROS module to set the boolean flag, this request is a code patch ready to merge and does not rely on the English language - in contrast to suggestions by you and dirk.

This is not as simple as just this package we're discussing how to change the overall behavior of the ROS environment, and this change cannot be made without changing all the other code paths that it effects as well as updating all the documentation that refers to how to setup environments. The "one line of code that has to be added to some osrf-ROS module" is not an accurate representation of the changes that need to be made. All the environment setup tools will need to be aware of this change and conditionally set the environment variable appropriately based on the use case.

>> and also ends up with a hardcoded path in the code that cannot be adapted for other use cases.

> To reiterate: POSIX conform installation should not be a special case of a catkin installation.
> The File System Hierarchy is not some random path. It's a very old standard (in the current version published in 2004 and catkin uses it for system versions of cmake, Python, Qt, etc. You should not adapt the required code "for other use cases".
> 
> Please consider merging this to support standard-conform package installation.
> Improving the code afterwards is definitely possible afterward.

We agree that conforming to standards is valuable and as you've pointed out we are well aware of it and do our best to support it. I did not say that `/usr/share` is a random path. However it is a path that is only valid in a subset of use cases. Off the top of my head this hard coded value is not appropriate for someone who wants to compile from source, who would install to `/usr/local/share` the same as with someone using homebrew, or someone using macports and installing to `/opt/local`. Similarly hard coding this value is not appropriate for supporting the installation of multiple distros side by side, which is our current default development environment. Thus as currently proposed this is hard coding a path which is not valid for many of our users use cases including our default installations such that you will need to work around it with other tools.

Please note that our usage of `/opt/ros` is also close to FHS compliant as we're [a registered provider with LANANA](http://www.lanana.org/lsbreg/providers/providers.txt). There are some issues within our use of the prefix that are not exactly FHS compliant such as our use of $PREFIX/etc but those are background tasks we're aware of but have not had the time to restructure our toolchain to resolve the biggest hurdles.

The main issue here is actually Debian specific, as it's [Debian Policy 9.9](https://www.debian.org/doc/debian-policy/ch-opersys.html#s9.9) that we're discussing. This is not part of FHS or POSIX, and I don't see an equivalent section in  the [Fedora Packaging Guidelines](https://fedoraproject.org/wiki/Packaging:Guidelines).  Saying that we "should not" support use cases outside of `/usr/share` is a unproductively dogmatic.

I know that this is a lot of back and forth for what looks like a small change, but please work with us to iterate and find the best solution overall. There are many potential solutions and we want to make the best choice since there's a large cost to make a change at this low of a level. And although there's not a REP on this at the moment since the ROS_PACKAGE_PATH notably predates the REP process, it's actually good material for a REP since it's a system wide change in behavior. Either way before we merge this we will need to make coordinated pull-requests against all the effected toolchains and identify all the documentation that needs to be updated so we can make a coordinated release and push out the updated documentation.

v4hn:
On Thu, Aug 18, 2016 at 03:31:01PM -0700, Tully Foote wrote:
> This is not as simple as just this package we're discussing how to change the overall behavior of the ROS environment, and this change cannot be made without changing all the other code paths that it effects as well as updating all the documentation that refers to how to setup environments. The "one line of code that has to be added to some osrf-ROS module" is not an accurate representation of the changes that need to be made.

Could we try to collect all locations that will need updating with a patch like this one in this request?
Which locations do you have in mind?

> I did not say that `/usr/share` is a random path. However it is a path that is only valid in a subset of use cases.
> Off the top of my head this hard coded value is not appropriate for someone who wants to compile from source, who would install to `/usr/local/share` the same as with someone using homebrew, or someone using macports and installing to `/opt/local`.

I think this could be resolved by substituting the `ROS_SYSTEM_PACKAGE_PATH` that is added in this request by [`$XDG_DATA_DIRS`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html).
Do you agree that searching non-recursively through `$XDG_DATA_DIRS` (which is a freedesktop standard with a standardized default value) would be better?

> please work with us to iterate and find the best solution overall.

Ok, how do we proceed?

wjwwood:
We can catalogue places where we would need to add `/usr/share` to the RPP automatically, but in my opinion, that is not currently the goal of this PR (though it may have been originally). There are two options here, which are not mutually exclusive:

- get `ROS_PACKAGE_PATH_NO_DEFAULTS` code in `rospkg` (with this pr) and set it in `roslib` (debian can patch `roslib` to not set it and the current version of `roslib` debian already does not set it)
- change all tools that read the RPP directly to implicitly add `/usr/share` (or some defaults)

The first one is my goal right now, since it improves the quality of life for users of both the debian ROS packages and the ros.org ROS packages. And only needs to touch `rospkg` and `roslib`, though they need a coordinated release. Since it only affects `rospkg`, the documentation is that of a special feature for just `rospkg`, not a change to how all system tools work. To that end, perhaps a better name would be `ROSPKG_NO_DEFAULTS` (I know we've gone back and forth on the name too much already).

The second option is also possible, but requires a lot more work since it needs to be written into documentation with justification, probably as a REP. It's roll out probably needs to happen in the next version of ROS, since unlike the "first" option it changes expected behavior (the first option could also break things if people don't upgrade both `rospkg` and `roslib` at the same time, but nominally should work the same as it does now). And we need to touch at least a few more places that read the RPP directly, possibly many places.

So my suggestion on how to proceed is to pursue the short term solution with `ROS_PACKAGE_PATH_NO_DEFAULTS` in this pr, and move the ongoing discussion about always embedding `/usr/share` into the RPP to a new issue or even to discourse to get feedback from a wider audience.

If there is at least some general consensus that option 1 would be good in the short term and does not exclude the possibility of option 2 in the future, then I would pick this up and make any remaining changes to this pr, open a pr against roslib, and get the ball moving on this.

---

A note about `XDG_DATA_DIRS`, you cannot assume that it is a single directory (which may be fine, but this pr currently treats the `ROS_SYSTEM_PACKAGE_PATH` as a single directory). On my Xenial machine it is set to `XDG_DATA_DIRS=/usr/share/ubuntu:/usr/share/gnome:/usr/local/share/:/usr/share/:/var/lib/snapd/desktop`. It is not set on my OS X machine (Darwin BSD), nor are similar variables always set in my docker instances (it depends on the base image and what is installed). So it's fine to use the `XDG` variables, but we shouldn't rely on them always being set. In general I don't see a problem with automatically doing non-recursive on those directories though.

v4hn:
On Tue, Aug 30, 2016 at 04:30:06PM -0700, William Woodall wrote:
> So my suggestion on how to proceed is to pursue the short term solution with `ROS_PACKAGE_PATH_NO_DEFAULTS` in this pr, and move the ongoing discussion about always embedding `/usr/share` into the RPP [...] to discourse to get feedback from a wider audience.

+1

> A note about `XDG_DATA_DIRS`, you cannot assume that it is a single directory.

Yes. The pull request has to be adapted.

> It is not set on my OS X machine (Darwin BSD), nor are similar variables always set in my docker instances (it depends on the base image and what is installed).
> So it's fine to use the `XDG` variables, but we shouldn't rely on them always being set.

The Freedesktop specs define the default value of the variable as `/usr/local/share/:/usr/share/`, i.e. if it is not set, we have to set it to that value to obey the specification.

jspricke:
Sorry for the late reply, have been busy with other things.

* William Woodall <notifications@github.com> [2016-08-30 16:30]:
> We can catalogue places where we would need to add `/usr/share` to the RPP automatically, but in my opinion, that is not currently the goal of this PR (though it may have been originally).

I think this catalogue would be a good idea. Till now I've found and
patched:

- rospkg
- rospack

> The first one is my goal right now, since it improves the quality of life for users of both the debian ROS packages and the ros.org ROS packages. And only needs to touch `rospkg` and `roslib`, though they need a coordinated release. Since it only affects `rospkg`, the documentation is that of a special feature for just `rospkg`, not a change to how all system tools work. To that end, perhaps a better name would be `ROSPKG_NO_DEFAULTS` (I know we've gone back and forth on the name too much already).

I don't think adding such a temporary solution would be right. For once
because it would be really confusing if it changes later and second I
don't see a use case for this, but a lot of use cases for the pull
request (i.e. all other distros ;) ).

> A note about `XDG_DATA_DIRS`, you cannot assume that it is a single directory

I've added `/usr/local/share`.

wjwwood:
Ok, sorry to have dropped off on this issue. Please accept my apologies, I started this response several times, but never get a chance to finish it until now.

We (some people at OSRF) had a meeting about this back a few months ago (I haven't had the time until now to write up a response) and the group consensus was to not do the "short-term" fix I described in https://github.com/ros-infrastructure/rospkg/pull/102#issuecomment-243613838.

Which was (in summary) to add these default values (based on `XDG` variables) in `rospkg` with the option to disable or override them with an environment variable and then empty that default list in our of our core ROS packages, like `roslib`. That way the default behavior isn't changed, but there would be a one line change in `roslib` that the Debian packaging process could make to enable this new behavior.

The original goal of this was (in part) to avoid the issue of `roslaunch` not working because the non-Debian-patched `rospkg` was installed and users tried to use it with the upstream Debian ROS packages (i.e. not the .deb's from packages.ros.org). Because of the nature of this pitfall, the C++ equivalent (`rospack`) was not affected and therefore did not need modification upstream with a way to disable the defaults.

Instead we would like to avoid this issue in a different way: by creating Debian conflicts between our core Python packages (like `python-catkin-pkg`) and some of the upstream Debian ROS packages. This would prevent the users from accidentally installing the packages.ros.org Python packages at the same time as the upstream Debian packages, which is the current recommendation anyways:

> We strongly recommend against mixing ROS packages from upstream Debian/Ubuntu and from packages.ros.org on the same system.
>
> -- http://wiki.ros.org/UpstreamPackages

--

To be clear, we are not closing the door on having these default values in the `ROS_PACKAGE_PATH`, but what we've decided not to do is to only add them to `rospkg` with an overriding mechanism like this. We may still take this pr more or less as-is, but we want to do so when we are ready to change the defaults and behavior of the `ROS_PACKAGE_PATH` is all places (`rospkg`, `rospack`, and others). And instead we'll avoid the user pitfall described above using conflicting packages to prevent the preconditions for the pitfall to occur.

dirk-thomas:
I will close the PR for now based on the last comment, the time of inactivity since then, and the fact that we have added the conflict rules to the Python packages deployed in the ROS apt repo since then.

Please feel free to comment and the PR can be reopened if necessary.

PR #104, opened 2016-08-10
Adding common SROS paths

ruffsl:
The SROS version of ros_comm used this to find expected SROS related paths

### DIFF
--- a/src/rospkg/environment.py
+++ b/src/rospkg/environment.py
@@ -42,6 +42,7 @@ import os
 ROS_ROOT         = "ROS_ROOT"
 ROS_PACKAGE_PATH = "ROS_PACKAGE_PATH"
 ROS_HOME         = "ROS_HOME"
+SROS_HOME        = "SROS_HOME"
 
 # override directory path to /etc/ros
 ROS_ETC_DIR      = "ROS_ETC_DIR"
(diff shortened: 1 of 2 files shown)


src/rospkg/environment.py
```diff
@@ -116,7 +121,7 @@ def get_ros_home(env=None):
     Get directory location of '.ros' directory (aka ROS home).
     possible locations for this. The :envvar:`ROS_HOME` environment
     variable has priority. If :envvar:`ROS_HOME` is not set,
-    ``$HOME/.ros/log`` is used.
```
dirk-thomas: Please provide a separate PR for changes unrelated to SROS.

dirk-thomas:
While the patch itself looks fine I would like see and understand the big picture how SROS should be integrated into the existing code base before considering to merge it.

src/rospkg/environment.py
ruffsl: Please see #105

ruffsl:
Most of the current work in SROS resides modifications to ros_comm. However the alterations here is where I thought to follow the convention to found the logic for finding ROS related system locations. You can see the list of SROS related environment variables at http://wiki.ros.org/SROS/EnvironmentVariables

The main purpose of this PR is to relocate the current state of SROS codebase under a more official  `sros` branch location (rather than my fork) so that I may finish writing more stable tutorials and development documentation. I've asked @tfoote to start an [`sros`](https://github.com/ros/ros_comm/tree/sros) branch for ros_comm my next PR there as well. These will serve to consolidate all SROS related progress to their originating codebases, rather than scattered between me and @codebot 's personal forks.

tfoote:
lgtm for merge into the sros branch. There will need to be a large discussion about how best to merge it back to the mainline. Clearly we're exploring an area we didn't design for.

dirk-thomas:
Is the patch in this PR supposed to be reviewed before merging or will we defer this until the branch is considered to be merged into *master*? If the latter please go ahead with the merge anytime.

PR #106, opened 2016-08-29
support and tests for slackware linux

nikonikolov:
Added support for Slackware Linux and tested it as required. See attached file for the log from testing. 
[rospkg_slackware_log.txt](https://github.com/ros-infrastructure/rospkg/files/443009/rospkg_slackware_log.txt)


tfoote:
Thanks for the contribution.

PR #109, opened 2016-09-16
add detection for Neon Linux OS

### LINKED ISSUE #108: Fix detection for derivative (rebranded) OS
for example Neon (Ubuntu), ...

As per specification, the `os-release` has `ID` and `ID_LIKE` fields.

Proposal:
Parse the `ID_LIKE` if the first try fails.

Reported in: ros/rosdistro#12716

breznak:
which is a (close) derivative of Ubuntu (LSB)

This alone seems to fix the problem. 

Fixes #108

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -642,6 +642,7 @@ OS_FUNTOO='funtoo'
 OS_GENTOO='gentoo'
 OS_LINARO='linaro'
 OS_MINT='mint'
+OS_NEON='neon'
 OS_OPENSUSE='opensuse'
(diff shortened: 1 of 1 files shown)


tfoote:
Thanks for the patch!

smihael:
This needs to be updated, as ID_LIKE has been changed to

    ID_LIKE="ubuntu debian"

in recent neon releases (tested with KDE neon User Edition 5.16).

tfoote:
Thanks @smihael for the notice. Could you test and propose a change in a PR to fix this? Or at least open an issue, otherwise your comment on this old PR will likely not be seen again.

PR #111, opened 2016-09-30
metadata and build script tweaks so it will build with modern catkin

codebot:
These tweaks to the CMakeLists.txt file and adding a minimal package.xml allow rospkg to build using "catkin build" which is convenient when you want to tinker with rospkg without messing up the system-installed version. :chicken:   I'm just proposing this to go on the sros branch for now, since I don't really understand all the implications of this tinkering.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,11 +1,11 @@
 cmake_minimum_required(VERSION 2.8.0)
 project(rospkg)
-find_package(catkin)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
I am fine with the change to go to the `sros` branch but I would argue that it should *not* be merged to `master`. So at some point you might want to look at a different approach.

PR #112, opened 2016-12-23
update to use new backup of old release repo on sourceforge.

tfoote:
Re: https://github.com/ros/rosdistro/issues/7484


dirk-thomas:
@tfoote What is the status of this ticket?

tfoote:
Sorry I missed the test failures and had the local changes uncommitted in my working copy. Should be good to go now.

PR #115, opened 2017-04-10
Support for Manjaro Linux

jaume-gasa:
Added support for Manjaro Linux (distro based on Arch Linux)

https://manjaro.org

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -398,6 +398,14 @@ class Arch(OsDetector):
         raise OsNotDetected('called in incorrect OS')
 
 
+class Manjaro(Arch):
+    """
+    Detect Manjaro.
+    """
+    def __init__(self, release_file='/etc/manjaro-release'):
+        super(Manjaro, self).__init__(release_file)
+
+
 class Centos(OsDetector):
     """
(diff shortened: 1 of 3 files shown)


wjwwood:
See: https://github.com/ros-infrastructure/rosdep/issues/507

jaume-gasa:
@wjwwood  I was able to run ```rosdep update``` correctly (https://imgur.com/a/JrJwu) and without the error I posted in ros-infrastructure/rosdep#507

wjwwood:
Great, thanks for the confirmation.

wjwwood:
I'm not sure when this change will be released (that's up to @dirk-thomas to decide when) but that would include `python-rospkg` (ubuntu) and `rospkg` in `pip` getting updated.  Separately, I don't know when the arch equivalent package would get updated.

dirk-thomas:
@wjwwood There is no reason anyone else can do releases of this package.

Since there are no other pending patches this should be released now (rather than wait for an undefined amount of time). I just released version 1.1.1.

wjwwood:
I just figured that we should batch them up, which I would leave to you. I wasn't thinking to release just for this patch, but that's fine too.

PR #116, opened 2017-04-18
Fix os_detect to correctly detect Elementary OS

Tidrek:
In python this is what `platform.linux_distribution()` returns:
`('"elementary"', '0.4', 'loki')`

People currently overcome it with setting manually `ROS_OS_OVERRIDE` environment variable, but it should normally work out of the box


wjwwood:
@Tidrek to be clear, I'll wait to merge this until you can comment on my question about older versions and whether or not it is reasonable to try and support the old and new identifier strings.

Tidrek:
I can confirm that people have the same problem with Luna 0.2 (ubuntu 14.04) and Freya 0.3 (ubuntu 12.04), but I don't know about the very first release of elementary. I'm pretty sure though, that the majority of eOS users run either Freya or Loki releases.

tfoote:
If they are EOL we can drop support, otherwise I'd like to extend this to support both formats. Clearly the current state was required for some older version.

Can you point to documentation about their state of support? I found one reference that suggested Elementary only supports the current release but that was just a forum post.

Tidrek:
I couldn't find official confirmation about EOL of older releases, even though I'm sure a small team of enthusiasts can't afford supporting anything older than 5 years. So, I downloaded all previous releases of elementary OS and ran them in live mode in VirtualBox. Below are the results:

jupiter
luna
freya

So, yeah, it was different before, but apparently with quotes, so it seems like os_detect.py never detected eOS correctly starting from Luna.

PR #117, opened 2017-07-13
set default Debian-Version

dirk-thomas:
Requires ros-infrastructure/ros_release_python#20.


dirk-thomas:
The CI failure is unrelated so I will move forward and merge this.

PR #121, opened 2017-08-23
os_detect: add support for openSUSE Leap 42.x and Tumbleweed

huangqinjin:
`/etc/SuSE-release` has been deprecated since openSUSE Leap 42.x.
```
~$ cat /etc/SUSE-brand 
openSUSE
VERSION = 42.1
~$ cat /etc/SuSE-release 
openSUSE 42.3 (x86_64)
VERSION = 42.3
CODENAME = Malachite
# /etc/SuSE-release is deprecated and will be removed in the future, use /etc/os-release instead
~$ cat /etc/os-release 
NAME="openSUSE Leap"
VERSION="42.3"
ID=opensuse
ID_LIKE="suse"
VERSION_ID="42.3"
PRETTY_NAME="openSUSE Leap 42.3"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:42.3"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
```

 openSUSE Tumbleweed has removed `/etc/SuSE-release` .
```
~$ cat /etc/SUSE-brand
openSUSE
VERSION = 13.3
~$ cat /etc/os-release 
NAME="openSUSE Tumbleweed"
# VERSION="20170810"
ID=opensuse
ID_LIKE="suse"
VERSION_ID="20170810"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:tumbleweed:20170810"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
```

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -616,6 +616,7 @@ class OsDetect:
                     self._os_version = os_detector.get_version()
                     self._os_codename = os_detector.get_codename()
                     self._os_detector = os_detector
+                    break
 
         if self._os_name:
             return self._os_name, self._os_version, self._os_codename
@@ -704,6 +705,7 @@ OsDetect.register_default(OS_MINT, LsbDetect("LinuxMint"))
 OsDetect.register_default(OS_NEON, LsbDetect("neon"))
 OsDetect.register_default(OS_OPENSUSE, OpenSuse())
 OsDetect.register_default(OS_OPENSUSE13, OpenSuse(brand_file='/etc/SUSE-brand'))
+OsDetect.register_default(OS_OPENSUSE, FdoDetect("opensuse"))
 OsDetect.register_default(OS_OSX, OSX())
 OsDetect.register_default(OS_QNX, QNX())
 OsDetect.register_default(OS_RHEL, Rhel())


tfoote:
This looks reasonable. However it's registering with the same key, which I think means that it's overriding the previous key. Will this method work for older OpenSUSE platforms? If so we should replace the old detection method. If not we will need to integrate this logic into the OpenSuse detection method.

huangqinjin:
It doesn't override. All detectors are in a list in the reverse order of registering. So, for opensuse, it checks the existence of `/etc/os-release` first, then of `/etc/SUSE-brand` and of `/etc/SuSE-brand` last. I break the loop in `detect_os()` after first detection succeeded.

gavanderhoorn:
Just leaving a note: tested this on (a Docker image of) OpenSUSE 42.3 and it seems to work. Running `os_detect.py` I get:

```
OS Name:     opensuse
OS Version:  42.3
OS Codename: 42.3
```

slightly off-topic: @huangqinjin: trying to build Kinetic from sources on `42.3`, did you have to do anything special? I can't get `rosdep` to see that I already have `libconsole_bridge0` installed.

huangqinjin:
@gavanderhoorn test if good for you.
```
function make_package() {
cat >src/$1/package.xml <<EOL
<?xml version="1.0"?>
<package>
  <name>$1</name>
  <version>0.0.0</version>
  <description>auto-generated</description>
  <maintainer email="nobody@nobody.com">nobody</maintainer>
  <license>BSD</license>
  <build_depend>$2</build_depend>
  <export>
    <build_type>cmake</build_type>
  </export>
</package>
EOL

depend="lib${1//[_-]/[_-]}-dev"  # original ubuntu package
find src -name package.xml -exec sed -i "s/$depend/$1/g" {} \;
}

git clone https://github.com/ros/console_bridge.git src/console_bridge && make_package console_bridge
git clone https://github.com/ros/urdfdom_headers.git src/urdfdom_headers && make_package urdfdom_headers
git clone https://github.com/ros/urdfdom.git src/urdfdom && make_package urdfdom urdfdom_headers
# https://github.com/ros/geometry2.git 0.5.16 doesn't support python3, use newer version instead

src/catkin/bin/catkin_make_isolated --install-space /opt/ros/kinetic --install --use-ninja -DCMAKE_BUILD_TYPE=Release

vi /opt/ros/kinetic/_setup_util.py
ENV_VAR_SUBFOLDERS = {
    'CMAKE_PREFIX_PATH': '',
    'LD_LIBRARY_PATH' if not IS_DARWIN else 'DYLD_LIBRARY_PATH': ['lib', 'lib64'],
    'PATH': 'bin',
    'PKG_CONFIG_PATH': [os.path.join('lib', 'pkgconfig'), os.path.join('lib64', 'pkgconfig')],
    'PYTHONPATH': ['lib/python3.4/site-packages', 'lib64/python3.4/site-packages'],
}
```

gavanderhoorn:
Thanks for the hints, I'll try them out a little later.

In summary:

 1. you do have to change the names of the `rosdep` keys for OpenSUSE
 1. `console_bridge` needs to be installed from source (I also did that)
 1. `urdf` and infrastructure also needs to be built from sources
 1. and we need to add/change (?) something to/in `/opt/ros/kinetic/_setup_util.py`?

gavanderhoorn:
re: 1: should probably contribute some `rosdep` rules for that.

huangqinjin:
@gavanderhoorn I didn't use `rosdep` to resolve dependencies. After downloaded all packages by `wstool init` and `wstool update`, I just started `src/catkin/bin/catkin_make_isolated` to see which dependencies are missing. Changing the depend keys is just for the topological order of catkin building.

For 64bit openSUSE, `CMAKE_INSTALL_LIBDIR` defined as `lib64` in `GNUInstallDirs` and `get_python_lib(plat_specific=True, prefix='')` returns `lib64/python3.4/site-packages`. But `CATKIN_GLOBAL_LIB_DESTINATION` and `CATKIN_PACKAGE_LIB_DESTINATION` used by catkin are hard-coded to `lib`. So after successfully built, I need modify `ENV_VAR_SUBFOLDERS`. For me, `console_bridge`, `urdfdom`, `urdfdom_headers` and `PyKDL` are located in `lib64`.

tfoote:
>  I can't get rosdep to see that I already have libconsole_bridge0 installed.

A quick workaround is to tell rosdep to `--skip-keys`. Alternatively you can also add local rosdep rules to set the resolution of those keys to `[]`

> Changing the depend keys is just for the topological order of catkin building.

Can you explain this a little bit more? WIth the current dependencies it should build in the correct order. 

Also note that the standard way to install things from source would be to run the default build of the tool and install it into the install space and not to try to build it inside the catkin workspace.

gavanderhoorn:
@tfoote wrote:
> @gavanderhoorn wrote:
> > I can't get rosdep to see that I already have libconsole_bridge0 installed.
> > 
> A quick workaround is to tell rosdep to `--skip-keys`. Alternatively you can also add local rosdep rules to set the resolution of those keys to `[]`

Yeah, I've used `skip-keys` for quite a few dependencies (mostly the `python-` ones), but couldn't find the right key for `libconsole_bridge0`. Asking it to ignore `console_bridge` didn't work for some reason.

tfoote:
The key for `console_bridge` is [`libconsole-bridge-dev`](https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L1456)

gavanderhoorn:
Right. That worked. Thanks for that.

huangqinjin:
I added `console_bridge` source to catkin workspace, but it wasn't built before something depend on it. It works if I build it alone and install into install-sapce, thanks.

gavanderhoorn:
@huangqinjin wrote:
> I added `console_bridge` source to catkin workspace, but it wasn't built before something depend on it. It works if I build it alone and install into install-sapce, thanks.

Recognising it as a system dependency I just built it 'manually' and installed it system-wide. `zypper` doesn't know about it, hence the `skip-keys`, but other than that it's working.

gavanderhoorn:
@tfoote: sorry for hijacking the thread.

The changes proposed by @huangqinjin seem to work: I can successfully use `os_detect.py` to get the proper output on recent OpenSUSE releases (I only tested on `42.3` (Leap) though).

Don't let me hold up the merge.

gavanderhoorn:
@huangqinjin: thanks for submitting the PR. I was able to build `ros_comm` without any major changes or work-arounds on Leap partly thanks to your contribution.

I've documented it [here](https://github.com/gavanderhoorn/docker_kinetic_on_opensuse_leap).

PR #122, opened 2017-09-26
os_detect: Added support for macos high sierra

smnogar:


wjwwood:
I just released `1.1.3` with this change.

PR #128, opened 2018-02-06
[rosversion] Option to remove new line at the end of the line.

130s:
Partially addresses https://github.com/ros/ros_comm/issues/1064

```
RUMER@paella$ ./scripts/rosversion
Usage:
rosversion <package/stack> [option]
rosversion -d [option]

Option:
-n: Add no new line at the end of the output line.

RUMER@paella$ ./scripts/rosversion -d
kinetic
RUMER@paella$ ./scripts/rosversion -d -n
kineticRUMER@paella$ ./scripts/rosversion roslaunch
1.12.12
RUMER@paella$ ./scripts/rosversion roslaunch -n
1.12.12RUMER@paella$
```

### DIFF
--- a/scripts/rosversion
+++ b/scripts/rosversion
@@ -29,6 +29,7 @@
 
 from __future__ import print_function
 
+import argparse
 import os
 import sys
 import traceback
@@ -37,7 +38,7 @@ import rospkg
 from rospkg.common import PACKAGE_FILE
 from rospkg.rospack import ManifestManager
 
-USAGE = "Usage: rosversion <package/stack> or rosversion -d"
+USAGE = """Returns the version of the given package, or the distribution version of ROS."""
 
 
 # for < fuerte, retrieve from roscore file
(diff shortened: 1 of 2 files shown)


scripts/rosversion
```diff
@@ -68,18 +74,31 @@ def get_distro_name_from_roscore():
...
 
-if len(sys.argv) == 2:
+def get_version_printer(nonewline=False):
+    """
+    @return: A method object to print the version string.
+    """
+    if nonewline:
+        return sys.stderr.write
+    else:
+        return print
```
dirk-thomas: Please use the keyword argument `end` of the print function for this.

scripts/rosversion
```diff
@@ -68,18 +74,31 @@ def get_distro_name_from_roscore():
...
+
+if len(sys.argv) >= 2:
+    version_printer = print
     stack_name = sys.argv[1]
 
+    if len(sys.argv) == 3 and sys.argv[2] == "-n":
```
dirk-thomas: Instead of starting to manually handle new arguments with custom logic please use `argparse` for it.

scripts/rosversion
```diff
@@ -37,7 +37,13 @@ import rospkg
 from rospkg.common import PACKAGE_FILE
 from rospkg.rospack import ManifestManager
 
-USAGE = "Usage: rosversion <package/stack> or rosversion -d"
+USAGE = """Usage:
+rosversion <package/stack> [option]
+rosversion -d [option]
+
+Option:
+-n: Add no new line at the end of the output line.
```
dirk-thomas: The help text is a bit difficult to read. I would suggest:

> Skip trailing new line

And based on that name the option `-s`.

scripts/rosversion
130s: I agree to use `argparse` or something organized argument parser. But because 2 usecases of `rosversion` where both takes a positional args, one of which starts with hyphen so that `argparse` automatically recognizes it as an optional, I can't think of how to implement using `argparse`. I'll keep thinking but if you can share your idea if any I'd appreciate.

```
rosversion <package/stack> 
```
or
```
rosversion -d                                                                                                                                                 
```

130s: [Asked on SO](https://stackoverflow.com/questions/48798761/positional-arg-starts-with-hyphen-in-argparse).

dirk-thomas:
```
parser = argparse.ArgumentParser()
group = parser.add_mutually_exclusive_group()
group.add_argument('package', nargs='?')
group.add_argument('-d', action='store_true')
```

130s:
@dirk-thomas thanks for the review.

Addressed all comments. Please re-review.

dirk-thomas:
In the `setup.py` file `argparse` needs to be added to `install_requires` (in alphabetical order).

In the `stdeb.cfg` file `python-argparse` has to be added to the `Depends` key in the `rospkg` section.

scripts/rosversion
```diff
@@ -68,47 +69,75 @@ def get_distro_name_from_roscore():
...
-            distro_name = os.environ['ROS_DISTRO']
-        else:
-            distro_name = get_distro_name_from_roscore()
-        if not distro_name:
-            print('<unknown>')
-        else:
-            print(distro_name)
-        sys.exit(0)
+parser = argparse.ArgumentParser(description=USAGE)
+if len(sys.argv) < 2:
```
dirk-thomas: This check is not valid. When the user e.g. passes `-s` the condition is `True` without having passed any of the required arguments.

I changed the patch to pass `required=True` to the `add_mutually_exclusive_group()` call.

scripts/rosversion
```diff
@@ -37,7 +38,7 @@ import rospkg
 from rospkg.common import PACKAGE_FILE
 from rospkg.rospack import ManifestManager
 
-USAGE = "Usage: rosversion <package/stack> or rosversion -d"
+USAGE = """Returns the version of the given package, or the distribution version of ROS."""
```
dirk-thomas: Since this is not a usage line anymore I have moved the string directly to the `ArgumentParser` constructor.

scripts/rosversion
```diff
@@ -68,47 +69,75 @@ def get_distro_name_from_roscore():
...
+parser = argparse.ArgumentParser(description=USAGE)
+if len(sys.argv) < 2:
+    parser.print_usage()
+    sys.exit(1)
+group = parser.add_mutually_exclusive_group()
+group.add_argument(
+    'package',
+    help="Package name. This name must be in ROS format, not OS package format. "
+         "E.g. CORRECT: roscpp, INCORRECT ros-kinetic-roscpp. See ROS naming "
+         "convention http://wiki.ros.org/ROS/Patterns/Conventions#Packages",
```
dirk-thomas: I don't think this level of verbosity is necessary here. I boiled it down to mention `roscpp` as an example which should be clear enough.

scripts/rosversion
```diff
@@ -68,47 +69,75 @@ def get_distro_name_from_roscore():
...
+    nargs='?')
+group.add_argument(
+    '-d', '--distro',
+    help="No argument needed. With this, ROS distribution will be returned.",
+    action='store_true')
+parser.add_argument(
+    "-s", "--skipnewline",
```
dirk-thomas: I added a dash for better readability: `--skip-newline`.

dirk-thomas:
This is how `argparse` renders it. I would think users will notice that when they actual try to pass a package name and `-d` and get the error message that it doesn't make sense together.

130s:
```
$ ./scripts/rosversion roscpp -d
usage: rosversion [-h] [-s] [-d] [package]
rosversion: error: argument -d/--distro: not allowed with argument package
```

Though I still think there's little harm to add clarification to help message, this error message gives enough clue. I'm ok with as is.

dirk-thomas:
Please feel free to add any kind of example you think helps.

Also please add the missing dependencies mention before.

130s:
Addressed your comment @dirk-thomas Please review.

Command output:
```
$ ./scripts/rosversion -h
usage: rosversion [-h] [-s] [-d] [package]

rosversion -d: Output the version of the given package
rosversion package: ROS distribution name

positional arguments:
  package             The ROS package name (e.g. 'roscpp')

optional arguments:
  -h, --help          show this help message and exit
  -s, --skip-newline  Skip trailing newline
  -d, --distro        Output the ROS distribution name
$ ./scripts/rosversion roscpp
1.12.12
$ ./scripts/rosversion roscpp -s
1.12.12$ ./scripts/rosversion roscpp -d
usage: rosversion [-h] [-s] [-d] [package]
rosversion: error: argument -d/--distro: not allowed with argument package
$ ./scripts/rosversion -d         
kinetic
$ ./scripts/rosversion -d -s
kinetic$ ./scripts/rosversion -d roscpp
usage: rosversion [-h] [-s] [-d] [package]
rosversion: error: argument package: not allowed with argument -d/--distro
$ 
```

130s:
@dirk-thomas could I get a review?

stdeb.cfg
```diff
@@ -1,7 +1,7 @@
 [rospkg]
 Debian-Version: 100
-Depends: python-catkin-pkg, python-rospkg-modules
-Depends3: python3-catkin-pkg, python3-rospkg-modules
+Depends: python-argparse, python-catkin-pkg, python-rospkg-modules
+Depends3: python-argparse, python3-catkin-pkg, python3-rospkg-modules
```
dirk-thomas: Please remove the change for Python 3. First it can't depend on the Python 2 version of argparse. Second `argparse` is part of Python 3 and ha no separate Debian package.

scripts/rosversion
```diff
@@ -68,47 +67,60 @@ def get_distro_name_from_roscore():
...
-                    root = ElementTree(None, package_manifest)
-                    version = root.findtext('version')
-                except Exception:
-                    pass
-        except rospkg.ResourceNotFound as e:
-            print("Cannot locate [%s]" % (stack_name))
-            sys.exit(1)
+parser = argparse.ArgumentParser(
+    description='rosversion -d: Output the version of the given package\n'
+    'rosversion package: ROS distribution name',
```
dirk-thomas: These two messages don't look consistent to me. The first one clearly says "Output the version of ...". The second one just says "ROS distribution name".

Please make the messages similar, e.g.:

> rosversion -d: Output the version of the given package
> rosversion package: Output the ROS distribution name

dirk-thomas: See 8dac7a11503f7f598fcb36125eff0f0f8d2f13d7.

stdeb.cfg
dirk-thomas: See d9f2760a21842f791ceabaa5c4cae8082fdc5968.

dirk-thomas:
Thank you for the patch.

130s:
Thank you for the last changes @dirk-thomas. Always learning a lot from your review.

Issue #131, opened 2018-02-09
[get_depends] ResourceNotFound exception occurs when "build_depend"-ed pkgs not found


130s:
Please correct me if I'm wrong, but it's legally possible that some packages don't get installed when you strictly install the packages only in dependency chain (e.g. using `rosdep install --from-paths src --ignore-src -r -y`). I've found those that are only `build_depend`-ed in the dependency chain might not get installed? (, an example of which in my case is `roslang`) I'm not sure exactly why, but I found it not installed on my environment (and it makes sense to me. <s>I guess this is related to why we needed version 3 of package.xml format</s>).

Then, `RosPack.get_depends` [traverses all `*_depend`](https://github.com/ros-infrastructure/rospkg/blob/8bfc04ae7062f7e4e4046f0175f60d2930f37f7a/src/rospkg/manifest.py#L419) existing in the `package.xml` format. This results in `ResourceNotFound` exception on the environment like I described above.

This is not intuitive so needs to be fixed IMO, but I'm not sure what's the best solution. Do we want to search only `run_depend` by default and add an option to search all `*_depend`s?
If someone guides the direction I might be able to create a patch.

```
$ apt-cache policy python-rospkg
python-rospkg:
  Installed: 1.1.4-100
  Candidate: 1.1.4-100
  Version table:
 *** 1.1.4-100 500
        500 http://packages.ros.org/ros/ubuntu xenial/main amd64 Packages
        500 http://packages.ros.org/ros/ubuntu xenial/main i386 Packages
        100 /var/lib/dpkg/status
     1.0.38-1 500
        500 http://us.archive.ubuntu.com/ubuntu xenial/universe amd64 Packages
        500 http://us.archive.ubuntu.com/ubuntu xenial/universe i386 Packages
$ env|grep -i pythonpath
PYTHONPATH=/opt/ros/kinetic/lib/python2.7/dist-packages
$ apt-cache policy ros-kinetic-roslang                                                                                     
ros-kinetic-roslang:
  Installed: (none)
  Candidate: 1.13.6-0xenial-20171104-174440-0800
  Version table:
     1.13.6-0xenial-20171104-174440-0800 500
        500 http://packages.ros.org/ros/ubuntu xenial/main amd64 Packages
$ ipython
Python 2.7.12 (default, Dec  4 2017, 14:50:18) 
Type "copyright", "credits" or "license" for more information.

IPython 2.4.1 -- An enhanced Interactive Python.

[...]

PR #129, opened 2018-02-08
Fix get_{depends, rosdeps} error when depended pkgs are not available in the environment

130s:
UPDATED issue description.

**Issue**

When some dependencies are not available on the environment, the following problems can occur for `RosPack.get_{depends, rosdeps}`:
- For a package available on your environment (e.g. `roscpp`), 1st invocation fails without printing the dependency. But 2nd invocation returns a list of dependencies.
- `get_depends`'s error message does not list all missing pkgs; in the example below, what's actually missing on the environment are `roslang` and `rosunit`, but it only shows rosunit. Each invocation lists only a single package missing.
- `get_rosdeps` returns an empty set (set()) on the 2nd invocation when there is more than one package unavailable on your environment (e.g. `foo` as reported originally at https://github.com/ros-infrastructure/rospkg/pull/129#issuecomment-364307544).

**Current output**

```
$ apt-cache policy python-rospkg
python-rospkg:
  Installed: 1.1.4-100
  Candidate: 1.1.4-100

$ env|grep -i pythonpath
PYTHONPATH=/opt/ros/kinetic/lib/python2.7/dist-packages
$ ipython
In [1]: from rospkg import RosPack
In [2]: rp = RosPack()
In [3]: rp.get_depends("roscpp")
---------------------------------------------------------------------------
ResourceNotFound                          Traceback (most recent call last)
<ipython-input-3-3996ba0273da> in <module>()
----> 1 rp.get_depends("roscpp")

/usr/lib/python2.7/dist-packages/rospkg/rospack.pyc in get_depends(self, name, implicit)
    238
    239             for p in names:
--> 240                 s.update(self.get_depends(p, implicit))
    241             # add in our own deps
    242             s.update(names)

/usr/lib/python2.7/dist-packages/rospkg/rospack.pyc in get_depends(self, name, implicit)
    238
    239             for p in names:
--> 240                 s.update(self.get_depends(p, implicit))
    241             # add in our own deps
    242             s.update(names)

/usr/lib/python2.7/dist-packages/rospkg/rospack.pyc in get_depends(self, name, implicit)
    232
    233             # take the union of all dependencies
--> 234             names = [p.name for p in self.get_manifest(name).depends]
    235
    236             # assign key before recursive call to prevent infinite case

/usr/lib/python2.7/dist-packages/rospkg/rospack.pyc in get_manifest(self, name)
    165             return self._manifests[name]
    166         else:
--> 167             return self._load_manifest(name)
    168
[...]

PR #132, opened 2018-02-09
[get_depends] Add option to specify the type of dependency.

130s:
Adding <s>`only_runtime_depend`</s> an option to specify the type of dependency to amend the situation in #131.

When the method is called multiple times with the same package name as an input, currently the resulted list of dependencies doesn't change because of the issues in `get_depends` that results in `get_depends` not finishing then incomplete dependencies. Fix for that is proposed in https://github.com/ros-infrastructure/rospkg/pull/129.

Output (UPDATE: This is old. See the output in newer comments)
<s>
```
$ ipython
:
In [1]: from rospkg import RosPack
In [2]: rp = RosPack()

In [7]: len(rp.get_depends("roscpp", only_runtime_depend=True))
Out[7]: 19

In [9]: len(rp.get_depends("roscpp", only_runtime_depend=False))
Out[9]: 19
```
</s>

### DIFF
--- a/src/rospkg/manifest.py
+++ b/src/rospkg/manifest.py
@@ -370,14 +370,17 @@ def _get_text(nodes):
 _static_rosdep_view = None
 
 
-def parse_manifest_file(dirpath, manifest_name, rospack=None):
+def parse_manifest_file(dirpath, manifest_name, rospack=None, typeof_depend=0):
     """
     Parse manifest file (package, stack).  Type will be inferred from manifest_name.
 
     :param dirpath: directory of manifest file, ``str``
     :param manifest_name: ``MANIFEST_FILE`` or ``STACK_FILE``, ``str``
     :param rospack: a RosPack instance to identify local packages as ROS packages
-
+    :param typeof_depend:
+            0: all kinds of depend.
+            1: only runtime_depend, such as run_depend in package.xml format 1 and 2.
+            2: only build_depend.
     :returns: return :class:`Manifest` instance, populated with parsed fields
     :raises: :exc:`InvalidManifest`
     :raises: :exc:`IOError`
@@ -416,7 +419,14 @@ def parse_manifest_file(dirpath, manifest_name, rospack=None):
             if _static_rosdep_view:
                 depends = set([])
                 rosdeps = set([])
-                for d in (p.buildtool_depends + p.build_depends + p.run_depends + p.test_depends):
+                depended_pkgs = []
+                if typeof_depend == 1:
+                    depended_pkgs = p.run_depends
+                elif typeof_depend == 2:
+                    depended_pkgs = p.build_depends
+                else:
+                    depended_pkgs = p.buildtool_depends + p.build_depends + p.run_depends + p.test_depends
+                for d in (depended_pkgs):
                     if (rospack and d.name in rospack.list()) or is_ros_package(_static_rosdep_view, d.name):
                         depends.add(d.name)
                     if is_system_dependency(_static_rosdep_view, d.name):
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -130,6 +130,9 @@ class ManifestManager(object):
         self._location_cache = None
         self._custom_cache = {}
 
+        # Status of the scope for getting dependencies.
+        self._typeof_dependency = 0  # Any type of depend
+
     @classmethod
     def get_instance(cls, ros_paths=None):
         """
@@ -157,14 +160,18 @@ class ManifestManager(object):
         return self._ros_paths[:]
     ros_paths = property(get_ros_paths, doc="Get ROS paths of this instance")
 
-    def get_manifest(self, name):
+    def get_manifest(self, name, typeof_depend=0):
         """
+        :param typeof_depend:
+                0: all kinds of depend.
+                1: only runtime_depend, such as run_depend in package.xml format 1 and 2.
+                2: only build_depend.
         :raises: :exc:`InvalidManifest`
         """
         if name in self._manifests:
             return self._manifests[name]
         else:
-            return self._load_manifest(name)
+            return self._load_manifest(name, typeof_depend)
 
     def _update_location_cache(self):
         global _cache_lock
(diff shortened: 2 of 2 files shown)


130s:
Please review.

Update from the original post:

- You can specify a few different types of dependencies for `get_depends`; `run_depend`, `build_depend`.
  - Returning only build-related dependency is useful for https://github.com/ros-infrastructure/rospkg/pull/133 (where you might want to see if there's direct build-dependency on the type of licenses you want to be careful with, e.g. static linking to L-GPL).
- When the dependency type passed is different from the one passed in the previous invocation, renew the cache.

```
$ ipython
:
In [1]: from rospkg import RosPack
In [2]: rp = RosPack()
In [3]: rp.get_depends("roscpp")
Out[3]:
['rosconsole',
 'catkin',
 'cmake_modules',
 'rospack',
 'rosbuild',
 'gencpp',
 'roslib',
 'geneus',
 'roscpp_serialization',
 'roscpp_traits',
 'message_generation',
 'roslang',
 'genlisp',
 'genpy',
 'std_msgs',
 'message_runtime',
 'rostime',
 'rosunit',
 'rosgraph_msgs',
 'genmsg',
 'xmlrpcpp',
 'gennodejs',
 'cpp_common']

In [4]: rp.get_depends?
Type:        instancemethod
String form: <bound method RosPack.get_depends of <rospkg.rospack.RosPack object at 0x7f33094527d0>>
File:        /home/n130s/ROS/ncws/src/ros-infrastructure/rospkg/src/rospkg/rospack.py
Definition:  rp.get_depends(self, name, implicit=True, typeof_depend=0)
Docstring:
Get dependencies of a resource.  If implicit is ``True``, this
includes implicit (recursive) dependencies.

:param name: resource name, ``str``
:param implicit: include implicit (recursive) dependencies, ``bool``
:param typeof_depend:
        0: all kinds of depend.
        1: only runtime_depend, such as run_depend in package.xml format 1 and 2.
        2: only build_depend.
:type typeof_depend: int
:returns: list of names of dependencies, ``[str]``
:raises: :exc:`InvalidManifest` If resource or any of its
  dependencies have an invalid manifest.

In [5]: rp.get_depends("roscpp", typeof_depend=1)
Out[5]:
['rosgraph_msgs',
 'genlisp',
 'genpy',
 'genmsg',
 'catkin',
 'rosconsole',
 'gencpp',
 'xmlrpcpp',
 'rostime',
 'message_runtime',
 'std_msgs',
 'geneus',
 'roscpp_serialization',
 'roscpp_traits',
 'gennodejs',
 'rosbuild',
 'message_generation',
 'cpp_common']

In [6]: rp.get_depends("roscpp", typeof_depend=2)
Out[6]:
['rosgraph_msgs',
 'rosconsole',
 'roslang',
 'xmlrpcpp',
 'std_msgs',
 'rostime',
 'roscpp_serialization',
 'roscpp_traits',
 'rosunit',
 'message_generation',
 'cpp_common']

In[7]: rp.get_depends("roscpp", implicit=False)
Out[7]:
['rosgraph_msgs',
 'rosconsole',
 'roslang',
 'catkin',
 'xmlrpcpp',
 'std_msgs',
 'message_runtime',
 'rostime',
 'roscpp_serialization',
 'roscpp_traits',
 'message_generation',
 'cpp_common']

In [8]: rp.get_depends("roscpp", implicit=False, typeof_depend=1)
Out[8]:
['rosgraph_msgs',
 'rosconsole',
 'xmlrpcpp',
 'std_msgs',
 'message_runtime',
 'rostime',
 'roscpp_serialization',
 'roscpp_traits',
 'cpp_common']

In [9]: rp.get_depends("roscpp", implicit=False, typeof_depend=2)
Out[9]:
['rosgraph_msgs',
 'rosconsole',
 'roslang',
 'xmlrpcpp',
 'std_msgs',
 'rostime',
 'roscpp_serialization',
 'roscpp_traits',
 'message_generation',
 'cpp_common']
```

src/rospkg/manifest.py
```diff
@@ -370,14 +370,17 @@ def _get_text(nodes):
...
     Parse manifest file (package, stack).  Type will be inferred from manifest_name.
 
     :param dirpath: directory of manifest file, ``str``
     :param manifest_name: ``MANIFEST_FILE`` or ``STACK_FILE``, ``str``
     :param rospack: a RosPack instance to identify local packages as ROS packages
-
+    :param typeof_depend:
+            0: all kinds of depend.
+            1: only runtime_depend, such as run_depend in package.xml format 1 and 2.
+            2: only build_depend.
```
dirk-thomas: Please do not use an opaque numeric value for this purpose. Instead pass an iterable with the names of the dependency types (similar to other ROS API, e.g. https://github.com/ros-infrastructure/rosdistro/blob/26c9dbfc9634b0dd1797509f981477f51b736839/src/rosdistro/dependency_walker.py#L66).

src/rospkg/rospack.py
```diff
@@ -157,14 +160,18 @@ def get_ros_paths(self):
         return self._ros_paths[:]
     ros_paths = property(get_ros_paths, doc="Get ROS paths of this instance")
 
-    def get_manifest(self, name):
+    def get_manifest(self, name, typeof_depend=0):
         """
+        :param typeof_depend:
+                0: all kinds of depend.
+                1: only runtime_depend, such as run_depend in package.xml format 1 and 2.
```
dirk-thomas: Format 2 doesn't specify `run_depend` anymore.

src/rospkg/rospack.py
```diff
@@ -204,40 +211,53 @@ def get_path(self, name):
...
+                2: only build_depend.
+        :type typeof_depend: int
         :returns: list of names of dependencies, ``[str]``
         :raises: :exc:`InvalidManifest` If resource or any of its
           dependencies have an invalid manifest.
         """
+        # When typeof_depend flag is different from the one previously used,
+        # flag, re-compute the dependency.
+        if not self._typeof_dependency == typeof_depend:
```
dirk-thomas: Please use `!=` instead of `not ==`.

src/rospkg/rospack.py
```diff
@@ -204,40 +211,53 @@ def get_path(self, name):
...
+            self._typeof_dependency = typeof_depend
+            # Resetting cache. This may not be necessary when the duplicate invocation
+            # issue is addressed https://github.com/ros-infrastructure/rospkg/pull/129#issuecomment-364572565
```
dirk-thomas: Can you please elaborate on this. How would the fix from 129 address this? Assuming 129 gets merged how do you expect this patch to change?

src/rospkg/rospack.py
```diff
@@ -157,14 +160,18 @@ def get_ros_paths(self):
...
+                0: all kinds of depend.
+                1: only runtime_depend, such as run_depend in package.xml format 1 and 2.
+                2: only build_depend.
         :raises: :exc:`InvalidManifest`
         """
         if name in self._manifests:
             return self._manifests[name]
         else:
-            return self._load_manifest(name)
+            return self._load_manifest(name, typeof_depend)
```
dirk-thomas: Please call keyword arguments explicitly using the keyword rather than as a positional argument.

Same below.

130s:
> I am not sure if rospkg is the right place to implement this feature. `rospkg` "by design" doesn't distinguish the different dependency types since it works with catkin package.xml files as well as rosbuild manifest.xml files.

But then I think we need to handle a situation like https://github.com/ros-infrastructure/rospkg/issues/131 where `get_depends`'s output is confusing due to `build_depend`-ed packages don't always get installed (it took me awhile to figure out why that happens as you see in the ticket).

A couple of thoughts:
- Add warning to `get_depends` to explain the situation when some packages aren't found.
- Or open a PR much like this one to somewhere else. Is `catkin_pkg` suitable?

dirk-thomas:
I don't like the idea that the `load_mainfest` function already applies custom logic based on the recursive dependency types. As seen in this patch it then require the cache to be invalidate when the dependency arguments change.

I think the following approach would be "better":

* `load_manifest` loads a manifest and provides all the necessary information atomically. Since different dependencies are necessary to distinguish them afterwards they would need to be returned separately.
* the resulting set of manifests can then be traversed using whatever kind of recursive dependency types are desired
* from the final set of packages the relevant information can be queried (e.g. the license as done in #133)

I haven't checked if `catkin_pkg` is more appropriate. I was just raising the concern that the current API in `rospkg` isn't sufficient (the manifest only has a generic `depends` attribute).

dirk-thomas:
I will go ahead and close this PR due to the long time of inactivity. Please feel free to comment and if necessary the ticket can be reopened.

PR #138, opened 2018-05-06
Python3 fix for manifest decoding

leolchat:
Doing the same kind of fixes which happened in the other ROS packages.

Without it I am getting
```
File "/usr/local/lib/python3.6/dist-packages/rospkg/manifest.py", line 434, in parse_manifest_file
    return parse_manifest(manifest_name, f.read(), filename)
  File "/usr/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 213: ordinal not in range(128)
```

### DIFF
--- a/src/rospkg/manifest.py
+++ b/src/rospkg/manifest.py
@@ -430,8 +430,8 @@ def parse_manifest_file(dirpath, manifest_name, rospack=None):
 
         return manifest
 
-    with open(filename, 'r') as f:
-        return parse_manifest(manifest_name, f.read(), filename)
+    with open(filename, 'rb') as f:
+        return parse_manifest(manifest_name, f.read().decode('utf-8'), filename)
 
 
 def parse_manifest(manifest_name, string, filename='string'):


dirk-thomas:
Please provide a reproducible example.

leolchat:
Well my current setup is a little complex, but basically:
- Start a bare ubuntu docker `docker pull ubuntu:18.04` then run the docker.
- Inside the docker by default the locale are setup to minimalistic POSIX
- Try to compile ROS from source with usual `rosinstall_generator`,`wstool`, then `rosdep`which is the one failing on some package manifest with non ASCII names.

leolchat:
PS forcing `LANG`to be some unicode fixes it, but I thought that this piece of code could follow the changes made previously https://github.com/ros-infrastructure/catkin_pkg/pull/178/files

dirk-thomas:
Thanks for the patch and the clarification. I am happy to merge this.

Buy trying to use ROS on a system without a proper locale set will likely result in many problems like that. So I would highly recommend to set a `utf-8` locale.

dirk-thomas:
This patch resulted in the regression described in #154.

PR #146, opened 2018-06-28
[doc] Update installation section

130s:
**Motivation**: Following can be the issue for some other pkgs as well, not just `rospkg`, but as a standalone pkg, it should still make sense to be documented within its own document.
- rosdep key (`python-rospkg`) and the pkg name is different. I noticed some people get confused how to install the binary. So clarifying the binary package name can help .
- I noticed some people don't fully utilize resolving dependency via rosdep (they end up not defining keys in `package.xml` and every time run e.g. `apt`).

### DIFF
--- a/doc/python_api.rst
+++ b/doc/python_api.rst
@@ -60,6 +60,28 @@ Common API
 Installation
 ------------
 
+Recommended automated installation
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Thanks for the update.

PR #147, opened 2018-06-28
Add macOS mojave to os_detect

tfoote:
Follow up: https://github.com/ros-infrastructure/rosdep/issues/615


130s:
Looks reasonable, but I leave decision up to maintainers.

PR #150, opened 2018-08-04
Use console_scripts for a portable way to generate rosversion

seanyen-msft:
In Windows, the shebang python script without file extension cannot be run directly. Either user needs to create a batch script as a shim or we can use console_scripts to generate portable rosversion (in Windows, it will be a executable shim.)

This pull request is to propose to use console_scripts.


dirk-thomas:
Thank you for this fix.

dirk-thomas:
:boom: see #153 for the fix...

PR #152, opened 2018-08-08
remove Ubuntu distros older than Trusty, remove Debian Wheezy

dirk-thomas:
Related to ros-infrastructure/catkin_pkg#232.

Basically the toolchain on Bionic embeds a Python dependency on version >= 2.7.5 which isn't available on Ubuntu Precise / Debian Wheezy. At the same time it removes other EOL Ubuntu distro before the last supported LTS which is currently Ubuntu Trusty.


dirk-thomas:
Reverted in 29d3a4b838ae4c78d8209a0f0e0b1574919d6cb6, released a patch release for Precise, reapplied in 8814ce7185474860860f28b2076bd50c7c5fb696.

Issue #154, opened 2018-08-08
InvalidManifest: [/opt/ros/kinetic/share/roslisp/manifest.xml] invalid XML


adrianheron:
Full error is:
```ERROR: Rosdep experienced an error: [/opt/ros/kinetic/share/roslisp/manifest.xml] invalid XML: 'ascii' codec can't encode character u'\xf6' in position 213: ordinal not in range(128)
Please go to the rosdep page [1] and file a bug report with the stack trace below.
[1] : http://www.ros.org/wiki/rosdep

rosdep version: 0.12.2

Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/main.py", line 140, in rosdep_main
    exit_code = _rosdep_main(args)
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/main.py", line 388, in _rosdep_main
    return _package_args_handler(command, parser, options, args)
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/main.py", line 484, in _package_args_handler
    return command_handlers[command](lookup, packages, options)
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/main.py", line 677, in command_install
    uninstalled, errors = lookup.resolve_all(packages, installer_context, implicit=options.recursive)
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/lookup.py", line 393, in resolve_all
    rosdep_keys = self.get_rosdeps(resource_name, implicit=implicit)
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/lookup.py", line 315, in get_rosdeps
    return self.loader.get_rosdeps(resource_name, implicit=implicit)
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/rospkg_loader.py", line 138, in get_rosdeps
    if resource_name in self.get_catkin_paths():
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/rospkg_loader.py", line 127, in get_catkin_paths
    self._catkin_packages_cache = dict(find_catkin_paths(self._rospack))
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/rospkg_loader.py", line 126, in find_catkin_paths
    filter(lambda x: src.get_manifest(x).is_catkin, src.list()))
  File "/usr/local/lib/python2.7/dist-packages/rosdep2/rospkg_loader.py", line 126, in <lambda>
[...]

PR #155, opened 2018-08-09
don't decode('utf-8') on python 2.x

HEllRZA:
fix for #154 in python2.x

### DIFF
--- a/src/rospkg/manifest.py
+++ b/src/rospkg/manifest.py
@@ -431,7 +431,11 @@ def parse_manifest_file(dirpath, manifest_name, rospack=None):
         return manifest
 
     with open(filename, 'rb') as f:
-        return parse_manifest(manifest_name, f.read().decode('utf-8'), filename)
+        if sys.version_info > (3, 0):
+            manifest_str = f.read().decode('utf-8')
+        else:
+            manifest_str = f.read()
+        return parse_manifest(manifest_name, manifest_str, filename)
 
 
 def parse_manifest(manifest_name, string, filename='string'):


src/rospkg/manifest.py
```diff
@@ -431,7 +431,11 @@ def parse_manifest_file(dirpath, manifest_name, rospack=None):
         return manifest
 
     with open(filename, 'rb') as f:
-        return parse_manifest(manifest_name, f.read().decode('utf-8'), filename)
+        if sys.version_info > (3, 0):
```
rsinnet: This is fine but I would probably do the following to remove repeated code
```
manifest_str = f.read()
if sys.version_info > (3, 0):
    manifest_str = manifest_str.decode('utf-8')
```

src/rospkg/manifest.py
```diff
@@ -432,7 +432,7 @@ def parse_manifest_file(dirpath, manifest_name, rospack=None):
 
     with open(filename, 'rb') as f:
         data = f.read()
-    if sys.version[0] >= 3:
+    if sys.version_info[0] >= 3:
```
rsinnet: :+1:

rsinnet:
@dirk-thomas, what are your thoughts on when this will make it into the PPA?

dirk-thomas:
We don't have a PPA for ROS packages. The Debian packages have already been released and should be available through the ROS apt repos.

rsinnet:
I see it now, great! Thanks!

dirk-thomas:
> I see it now, great! Thanks!

Thanks for confirming that it works.

PR #156, opened 2018-09-06
Add os-release fall back and Clear Linux distro.

forwardtape:
To integrate rospkg to Clear Linux it is need to improve the os_detect.py.

* A fall back location of os-release was added following
  the freedesktop os-release manual. 

* The entry of Clear Linux distro was added.

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -71,23 +71,25 @@ def read_issue(filename="/etc/issue"):
     return None
 
 
-def read_os_release(filename=""):
+def read_os_release(filename=None):
     """
-    :returns: Dictonary of key value pairs from /etc/os-release or fallback to /usr/lib/os-release, with quotes stripped from values
+    :returns: Dictionary of key value pairs from /etc/os-release or fallback to 
+      /usr/lib/os-release, with quotes stripped from values
     """
-    release_info = {}
-    if not filename:
-        filename = "/etc/os-release"
+    if filename is None:
+        filename = '/etc/os-release'
         if not os.path.exists(filename):
-            filename = "/usr/lib/os-release"
+            filename = '/usr/lib/os-release'
 
-    if os.path.exists(filename):
-        with codecs.open(filename, 'r', encoding=locale.getpreferredencoding()) as f:
-            for line in f:
-                key, val = line.rstrip('\n').partition('=')[::2]
-                release_info[key] = val.strip('"')
-            return release_info
-    return None
+    if not os.path.exists(filename):
+        return None
+
+    release_info = {}
+    with codecs.open(filename, 'r', encoding=locale.getpreferredencoding()) as f:
+        for line in f:
+            key, val = line.rstrip('\n').partition('=')[::2]
+            release_info[key] = val.strip('"')
+    return release_info
 
 
 class OsNotDetected(Exception):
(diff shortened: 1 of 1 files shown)


src/rospkg/os_detect.py
```diff
@@ -73,9 +73,11 @@ def read_issue(filename="/etc/issue"):
 
 def read_os_release(filename="/etc/os-release"):
     """
-    :returns: Dictonary of key value pairs from /etc/os-release, with quotes stripped from values
+    :returns: Dictonary of key value pairs from /etc/os-release or fallback to /usr/lib/os-release, with quotes stripped from values
```
tfoote: Can you not just pass `filename='/usr/lib/os-release'` to the `FdoDetector` for Clear Linux?

That's why we have the parameter for non-standard locations.

tfoote: Sorry after reading more in https://github.com/clearlinux/distribution/issues/180#issuecomment-419133986 , the fallback behavior is expected of os-release

src/rospkg/os_detect.py
```diff
@@ -695,6 +697,7 @@ def get_codename(self):
 OS_SLACKWARE = 'slackware'
 OS_UBUNTU = 'ubuntu'
 OS_WINDOWS = 'windows'
+OS_CLEARLINUX = 'Clear Linux OS for Intel Architecture'
```
tfoote: This would be better as a single tag like string not a full sentence with spaces etc.

src/rospkg/os_detect.py
```diff
@@ -73,9 +73,11 @@ def read_issue(filename="/etc/issue"):
 
 def read_os_release(filename="/etc/os-release"):
     """
-    :returns: Dictonary of key value pairs from /etc/os-release, with quotes stripped from values
+    :returns: Dictonary of key value pairs from /etc/os-release or fallback to /usr/lib/os-release, with quotes stripped from values
     """
     release_info = {}
+    if not os.path.exists(filename):
+        filename = "/usr/lib/os-release"
```
dirk-thomas: I find it weird that this fallback is being used even when a non-standard `filename` is passed to the function especially since the fallback can't be changed.

miguelinux: `/etc/os-releaes`and `/usr/lib/os-release` are standard locations as said in
 https://www.freedesktop.org/software/systemd/man/os-release.html. 

When a non-standard `filename` is passed, then it must exist as the programmer is telling where the file live.

dirk-thomas: > When a non-standard `filename` is passed, then it must exist as the programmer is telling where the file live.

That is true. I just wouldn't rely on it. Instead the default value of the argument could be a tuple instead and if the first item doesn't exist the next one is considered. That avoids falling back to the new location in case a custom `filename` is passed.

forwardtape: In that case the method could be written as following:
* Change as empty the input parameter `filename`.
* If there is no input string path parameter then:
  + Check both locations _(as man os-release indicates)_.
* Else: 
  + Check only the input string `filename`.

I think in this particular method it is not necessary the usage of tuples, and doing it this way it is avoided falling back to the new location in case a custom `filename` is passed.

forwardtape:
`OS_CLEARLINUX` has been updated as mentioned by @tfoote.
@dirk-thomas I updated os_release method, let me know what do you think.

dirk-thomas:
Looks good to me. I changed the default value to be `None` and shuffled the code slightly to be in three steps:

1. use default value if none is passed (falling back to second if first doesn't exist)
2. return `None` when the requested file doesn't exist
3. parse the file and return the key value pairs

Thank you for working on this enhancement.

PR #164, opened 2019-02-23
[manifest] Use list of str instead of a single str, to handle better packages with multiple licenses.

130s:
# Changes suggested

Add a `licenses` field to Manifest class (in addition to `license` field).

# Background

Although {manifest, package}.xml format allows to define separate `license` xml elements, rospkg handles multiple licenses as a single string joined by a comma, which is inconvenient depending on the consumers' usecase.

Pointed out at https://github.com/ros-infrastructure/rospkg/pull/133#discussion_r253230420

# Discussion
- If we want to maintain backward compatibility, `license` str field shouldn't change. So without making any change to it, this PR only adds a new field. This could cause a confusion.
  - One way to mitigate I thought was to print a deprecation warning for `license` field when it's called by a consumer. But because it looks like `license` is (and other fields are) intended to be accessed directly (without getter method/structure), I couldn't think of a way to do that.

# Dependency on other PRs/issues
None

# Test sample result
```
$ ipython
Python 2.7.12 (default, Nov 12 2018, 14:36:49)
Type "copyright", "credits" or "license" for more information.

IPython 2.4.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import rospkg
In [2]: rp = rospkg.RosPack()
In [4]: p = rp.get_manifest("rviz")
In [5]: p.license
Out[5]: 'BSD, Creative Commons'

In [6]: p.licenses
Out[6]: ['BSD', 'Creative Commons']
```

# Review items suggested
- [x] Requester provides reproducible commands and sample test result.
- [x] CI should pass.
- [ ] Code review.

### DIFF
--- a/test/test_rospkg_catkin_packages.py
+++ b/test/test_rospkg_catkin_packages.py
@@ -58,3 +58,11 @@ def test_get_manifest():
     manager = rospkg.rospack.ManifestManager(rospkg.common.MANIFEST_FILE, ros_paths=[search_path])
     manif = manager.get_manifest("foo")
     assert(manif.type == "package")
+
+
+def test_licenses():
+    search_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'catkin_package_tests'))
+    manager = rospkg.rospack.RosPack(ros_paths=[search_path])
+    manif = manager.get_manifest("foo")
+    assert(manif.license == "BSD, LGPL")
+    assert(len(manif.licenses) == 2)  # package.xml in 'foo' defines BSD and LGPL
--- a/test/test_rospkg_catkin_packages.py
+++ b/test/test_rospkg_catkin_packages.py
@@ -60,7 +60,11 @@ def test_get_manifest():
 
 
 def test_licenses():
-    manager = rospkg.rospack.RosPack(ros_paths=[search_path])
-    manif = manager.get_manifest("foo")
-    assert(manif.license == "BSD, LGPL")
-    assert(len(manif.licenses) == 2)  # package.xml in 'foo' defines BSD and LGPL
+    rospack = rospkg.rospack.RosPack(ros_paths=[search_path]) 
+    licenses_list = ["BSD", "LGPL"]
+    licenses_str = ", ".join(licenses_list)
+    manif = rospack.get_manifest("foo")
+    print("DEBUG: manif.licenses={}".format(manif.licenses))
+    assert(manif.license == licenses_str)
+    for license in manif.licenses:
+        assert(license in licenses_list)
--- a/test/test_rospkg_catkin_packages.py
+++ b/test/test_rospkg_catkin_packages.py
@@ -36,10 +36,10 @@ import os
 
 import rospkg
 
+search_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'catkin_package_tests'))
+
 
 def test_find_packages():
-    search_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'catkin_package_tests'))
-
     manager = rospkg.rospack.ManifestManager(rospkg.common.MANIFEST_FILE, ros_paths=[search_path])
     # for backward compatibility a wet package which is not a metapackage is found when searching for MANIFEST_FILE
     assert(len(manager.list()) == 1)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Please add a test covering the new attribute.

130s:
@dirk-thomas Thanks for the suggestion. Please re-review including the new test case.

test/test_rospkg_catkin_packages.py
```diff
@@ -54,7 +54,13 @@ def test_find_packages():
...
 
 def test_get_manifest():
-    search_path = os.path.abspath(os.path.join(os.path.dirname(__file__), 'catkin_package_tests'))
     manager = rospkg.rospack.ManifestManager(rospkg.common.MANIFEST_FILE, ros_paths=[search_path])
     manif = manager.get_manifest("foo")
     assert(manif.type == "package")
+
+
+def test_licenses():
+    manager = rospkg.rospack.RosPack(ros_paths=[search_path])
```
dirk-thomas: Please change this variable name to `rospack` since it is not a manager.

test/test_rospkg_catkin_packages.py
```diff
@@ -54,7 +54,13 @@ def test_find_packages():
...
+    manif = manager.get_manifest("foo")
+    assert(manif.license == "BSD, LGPL")
+    assert(len(manif.licenses) == 2)  # package.xml in 'foo' defines BSD and LGPL
```
dirk-thomas: Instead of mentioning it in the comment please check both license makes explicitly with `assert`.

130s:
@dirk-thomas Hope all your comments addressed now (I resolved them on GitHub. Should I have not done so?).

test/test_rospkg_catkin_packages.py
```diff
@@ -54,7 +54,17 @@ def test_find_packages():
...
     manif = manager.get_manifest("foo")
     assert(manif.type == "package")
+
+
+def test_licenses():
+    rospack = rospkg.rospack.RosPack(ros_paths=[search_path]) 
+    licenses_list = ["BSD", "LGPL"]
+    licenses_str = ", ".join(licenses_list)
+    manif = rospack.get_manifest("foo")
+    print("DEBUG: manif.licenses={}".format(manif.licenses))
```
dirk-thomas: This debug output should be removed.

test/test_rospkg_catkin_packages.py
```diff
@@ -54,7 +54,17 @@ def test_find_packages():
...
+    assert(manif.license == licenses_str)
```
dirk-thomas: Please avoid trivial single use variables and use `", ".join(licenses_list)` here directly.

test/test_rospkg_catkin_packages.py
```diff
@@ -54,7 +54,17 @@ def test_find_packages():
...
+    for license in manif.licenses:
```
dirk-thomas: `license` is a builtin in Python and should therefore not be used for a variable name.

130s: I wasn't aware of [license](https://stackoverflow.com/questions/21503045/what-is-the-keyword-license-used-for-in-python) constant...Updated.

test/test_rospkg_catkin_packages.py
dirk-thomas: This comment has not been addressed yet.

test/test_rospkg_catkin_packages.py
dirk-thomas: This comment has not been addressed yet.

dirk-thomas:
Please be more thorough when addressing the provided. The number of iterations this rather small patch takes is too high and at some point I might not be able to spend more time on reviewing / iterating on it.

dirk-thomas:
Looks good now. Thanks for the patch.

130s:
Thank you for the quick iteration @dirk-thomas! 

(Sorry for the crying-wolf update...I pushed to a wrong remote without knowing.)

PR #165, opened 2019-03-10
Correct typo in doc

yotabits:
ala instead of aka


dirk-thomas:
Thanks for the patch.

PR #166, opened 2019-03-30
Initial OpenEmbedded support for rospkg

andre-rosa:
* Is required by related changes in rosdep, rosdistro
    https://github.com/ros-infrastructure/rosdep/pull/673
    https://github.com/ros/rosdistro/pull/20763
  * Please see the proposal below:
    https://discourse.ros.org/t/a-proposal-for-a-superflore-oe-recipe-generation-scheme/8401

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -215,6 +215,24 @@ class FdoDetect(OsDetector):
         raise OsNotDetected("called in incorrect OS")
 
 
+class OpenEmbedded(OsDetector):
+    """
+    Detect OpenEmbedded.
+    """
+    def is_os(self):
+        return "ROS_OS_OVERRIDE" in os.environ and os.environ["ROS_OS_OVERRIDE"] == "openembedded"
+
+    def get_version(self):
+        if self.is_os():
+            return ""
+        raise OsNotDetected('called in incorrect OS')
+
+    def get_codename(self):
+        if self.is_os():
+            return ""
+        raise OsNotDetected('called in incorrect OS')
+
+
 class OpenSuse(OsDetector):
     """
     Detect OpenSuse OS.
@@ -693,6 +711,7 @@ OS_GENTOO = 'gentoo'
 OS_LINARO = 'linaro'
 OS_MINT = 'mint'
 OS_NEON = 'neon'
+OS_OPENEMBEDDED = 'openembedded'
 OS_OPENSUSE = 'opensuse'
 OS_TIZEN = 'tizen'
 OS_OPENSUSE13 = 'opensuse'
@@ -719,6 +738,7 @@ OsDetect.register_default(OS_GENTOO, Gentoo())
 OsDetect.register_default(OS_LINARO, LsbDetect("Linaro"))
 OsDetect.register_default(OS_MINT, LsbDetect("LinuxMint"))
 OsDetect.register_default(OS_NEON, LsbDetect("neon"))
+OsDetect.register_default(OS_OPENEMBEDDED, OpenEmbedded())
 OsDetect.register_default(OS_OPENSUSE, OpenSuse())
 OsDetect.register_default(OS_OPENSUSE13, OpenSuse(brand_file='/etc/SUSE-brand'))
 OsDetect.register_default(OS_OPENSUSE, FdoDetect("opensuse"))


src/rospkg/os_detect.py
```diff
@@ -215,6 +215,24 @@ def get_codename(self):
         raise OsNotDetected("called in incorrect OS")
 
 
+class OpenEmbedded(OsDetector):
+    """
+    Detect OpenEmbedded.
+    """
+    def is_os(self):
+        return "ROS_OS_OVERRIDE" in os.environ and os.environ["ROS_OS_OVERRIDE"] == "oe"
```
tfoote: Although this will work for the current use case of needing to support rosdep during the cross compile stage. I suspect that we can actually detect OpenEmbedded systems at runtime which will enable more use cases down the line. For example rosdep could be able to validate dependencies are available even if it can't install them on a running system.

Poking around I found that ther e does look to be an /etc/issue generated: http://cgit.openembedded.org/openembedded-core/tree/meta/recipes-core/base-files/base-files_3.0.14.bb?h=master

src/rospkg/os_detect.py
```diff
@@ -693,6 +711,7 @@ def get_codename(self):
 OS_LINARO = 'linaro'
 OS_MINT = 'mint'
 OS_NEON = 'neon'
+OS_OPENEMBEDDED = 'oe'
```
tfoote: As a side comment, `oe` is a little short and hard to understand if you don't know what you're looking at, but `openembedded` is really long. So I don't have a better suggestion.

src/rospkg/os_detect.py
```diff
@@ -215,6 +215,24 @@ def get_codename(self):
...
 
 
+class OpenEmbedded(OsDetector):
+    """
+    Detect OpenEmbedded.
+    """
+    def is_os(self):
+        return "ROS_OS_OVERRIDE" in os.environ and os.environ["ROS_OS_OVERRIDE"] == "oe"
+
+    def get_version(self):
```
tfoote: There is a DISTRO_VERSION that could be checked for in /etc/issue. This might be valuable to pass through if it's available.

src/rospkg/os_detect.py
andre-rosa: @tfoote << Many thanks for your review and insights!

Not sure if running rosdep on the target side makes sense _yet_.
As per your example, what would you do with this information? I guess you'd go to the build system and generate a new package or image containing the dependencies. Then why not run rosdep just on the build system?

issue.net on [poky-thud-20.0.1.tar.bz2](http://downloads.yoctoproject.org/releases/yocto/yocto-2.6.1/poky-thud-20.0.1.tar.bz2) is _"Poky (Yocto Project Reference Distro) 2.6.1 qemux86 /dev/ttyS0"_ . We may eventually decide on a way to detect OpenEmbedded but I'd prefer to use the override mechanism to avoid any misdetections and messing with other OSs for now.

src/rospkg/os_detect.py
andre-rosa: Given that we're not supporting running rosdep on the OpenEmbedded target system initially, it'll be better to add it later on, when we can actually test it.

There are multiple possibilities for running rosdep for OpenEmbedded:
  1. Run it on the host machine that cross-compiled;
  2. Run it on the target device that was cross-compiled;
  3. Run it on the host machine that cross-compiled but connect to the target device over ssh...

src/rospkg/os_detect.py
andre-rosa: Agree; will make it openembedded so that's explicit for everybody.

dirk-thomas:
@andre-rosa @tfoote What is the status of this PR?

andre-rosa:
@dirk-thomas @tfoote @allenh1 @cottsay @herb-kuta-lge @jb-seo << This PR works well for generating OpenEmbedded recipes by superflore when combined with its other dependent PRs:
https://github.com/ros-infrastructure/rosdep/pull/673
https://github.com/ros/rosdistro/pull/20763
https://github.com/ros-infrastructure/superflore/pull/168

**Current status:**

1. We successfully migrated [BeanBirdBot](https://github.com/lgsvl/build-ros2-lgsvl) from manually crafted ROS 2 recipes based on bouncy to automatically generated recipes by superflore and tested it on Rpi3 hardware;
2. We successfully build a basic roscore ROS 2 Crystal image with Yocto 2.6 Thud release and ros2cli works well when doing basic testing under qemu;
3. We're working on getting a full build of ros-world Crystal + Yocto Thud ([WIP patches](https://github.com/herb-kuta-lge/meta-ros2/commits/r.hk/thud-superflore-crystal));
4. We can now see _dashing_ published to the index. As soon as its [distribution cache](http://repo.ros2.org/rosdistro_cache/dashing-cache.yaml.gz) gets populated more we'll also proceed to build it.

PR #170, opened 2019-06-14
Add Mac OS 10.15: Catalina

rotu:
Add Mac OS 10.15: Catalina so the ROS2 build system doesn't crash when calling `rosdep install` on a Catalina system.


dirk-thomas:
Thanks for the patch.

PR #171, opened 2019-08-03
os_detect: add support for NixOS

lopsided98:
This is the first step in adding support for [Nix](https://nixos.org/) to ROS. The Nix package manager is often used with NixOS, but it runs on any Linux distribution as well as macOS. When packages are built, `/etc/os-release` is not available in the build sandbox, so `ROS_OS_OVERRIDE` has to be used anyway. In order to fit cleanly into the ROS infrastructure, I am pretending that Nix and NixOS have the traditional package manager and distribution relationship.

I am working on adding support for Nix to various ROS projects:
* https://github.com/lopsided98/rosdistro/tree/nixos-support
* https://github.com/lopsided98/rosdep/tree/nixos-support
* https://github.com/lopsided98/superflore/tree/nixos-support

I maintain a [Superflore generated nixpkgs overlay](https://github.com/lopsided98/nix-ros-overlay) and have successfully built over 3600 ROS packages on a [Hydra](https://nixos.org/hydra/) build server (not yet publicly accessible)

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -713,16 +713,16 @@ OS_MX = 'mx'
 OS_NEON = 'neon'
 OS_OPENEMBEDDED = 'openembedded'
 OS_OPENSUSE = 'opensuse'
-OS_TIZEN = 'tizen'
 OS_OPENSUSE13 = 'opensuse'
+OS_TIZEN = 'tizen'
 OS_OSX = 'osx'
 OS_QNX = 'qnx'
 OS_RHEL = 'rhel'
 OS_SLACKWARE = 'slackware'
 OS_UBUNTU = 'ubuntu'
-OS_WINDOWS = 'windows'
 OS_CLEARLINUX = 'clearlinux'
 OS_NIXOS = 'nixos'
+OS_WINDOWS = 'windows'
 
 OsDetect.register_default(OS_ALPINE, FdoDetect("alpine"))
 OsDetect.register_default(OS_ARCH, Arch())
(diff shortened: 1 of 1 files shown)


lopsided98:
It will succeed on NixOS, just not in the build sandbox. Its just that rosdep/rospkg will most commonly be used inside the build sandbox.

tfoote:
ahh, ok. nevermind then. LGTM

src/rospkg/os_detect.py
```diff
@@ -713,15 +713,16 @@ def get_codename(self):
 OS_NEON = 'neon'
 OS_OPENEMBEDDED = 'openembedded'
 OS_OPENSUSE = 'opensuse'
-OS_TIZEN = 'tizen'
 OS_OPENSUSE13 = 'opensuse'
+OS_TIZEN = 'tizen'
```
dirk-thomas: Please revert unrelated changes like this.

src/rospkg/os_detect.py
```diff
@@ -713,15 +713,16 @@ def get_codename(self):
...
 OS_OSX = 'osx'
 OS_QNX = 'qnx'
 OS_RHEL = 'rhel'
 OS_SLACKWARE = 'slackware'
 OS_UBUNTU = 'ubuntu'
-OS_WINDOWS = 'windows'
```
dirk-thomas: Same here.

src/rospkg/os_detect.py
lopsided98: I moved those changes to a separate commit. I can remove them if you want, but it just annoyed me that the definitions were in a different order than the registrations.

dirk-thomas:
Thanks for the patch.

PR #178, opened 2019-10-09
Strip Alpine's patch version from OS codename

### LINKED ISSUE #177: Simplify AplineOS codename (remove patch version)
In order to release ROS on Alpine Linux in future, it will be necessary to specify the Alpine versions in the relevant distribution file (e.g. [rosdistro/melodic/distribution.yaml](https://github.com/ros/rosdistro/blob/master/melodic/distribution.yaml)). Currently, Alpine's OS codename is formatted according to the semantic versioning system MAJOR.MINOR.PATCH. This means that it is also necessary to specify Alpine versions using this format inside distribution.yaml files. However, it seems unnecessary to match against the operating system's patch version.

An example of how we're dealing with the Alpine formatted release name names can be found [here](https://github.com/seqsense/aports-ros-experimental/blob/master/backports/py-rosdistro/force-to-enable-alpine-platform.patch)
In that instance, we are required to specify patch level versions like:
```
       if name == 'kinetic':
           self.release_platforms['alpine'].append('3.7.0')
           self.release_platforms['alpine'].append('3.7.1')
           self.release_platforms['alpine'].append('3.7.2')
           self.release_platforms['alpine'].append('3.7.3')
```
But, it is desirable to specify the version like so:
```
       if name == 'kinetic':
           self.release_platforms['alpine'].append('3.7')
```
While the example is just a patch, in future we hope that Alpine will be fully support by ROS, and therefore wish to avoid the overly verbose requirement of specifying patch level versions.

MarkHedleyJones:
Closes #177 

This pull request simplifies the Alpine Linux version string used when specifying ROS distribution OS support. It is accompanied by https://github.com/ros-infrastructure/rosdep/pull/716

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -214,6 +214,19 @@ class FdoDetect(OsDetector):
         raise OsNotDetected("called in incorrect OS")
 
 
+class Alpine(FdoDetect):
+    """
+    Detect Alpine Linux
+    """
+    def __init__(self):
+        super(Alpine, self).__init__('alpine')
+
+    def get_codename(self):
+        if self.is_os():
+            return ".".join(self.get_version().split('.')[:2])
+        raise OsNotDetected('called in incorrect OS')
+
+
 class OpenEmbedded(OsDetector):
     """
     Detect OpenEmbedded.
@@ -725,7 +738,7 @@ OS_CLEARLINUX = 'clearlinux'
 OS_NIXOS = 'nixos'
 OS_WINDOWS = 'windows'
 
-OsDetect.register_default(OS_ALPINE, FdoDetect("alpine"))
+OsDetect.register_default(OS_ALPINE, Alpine())
 OsDetect.register_default(OS_ARCH, Arch())
 OsDetect.register_default(OS_MANJARO, Manjaro())
 OsDetect.register_default(OS_CENTOS, Centos())


cottsay:
I did make this [same kind of change for RHEL](https://github.com/ros-infrastructure/rosdep/pull/667), but it was made in rosdep itself. My thought was that the value used to identify in rosdep is what needed to change, and there was no need to decrease the amount of information that rospkg was currently exposing.

Just a thought.

MarkHedleyJones:
Nice, I like the simplicity of that. But wouldn't the inconsistency between rosdep and rospkg become an issue at some point?

For example (when modifying only rosdep as you did), when running `rospkg/os_detect.py` on Alpine I am presented with:
```
OS Name:     alpine
OS Version:  3.7.3
OS Codename: 3.7.3
```
But `rosdep db` outputs
```
OS NAME: alpine
OS VERSION: 3.7
```
So it's apparent that something funny is going on there.

With this pull request, `os_detect.py` yields:
```
OS Name:     alpine
OS Version:  3.7.3
OS Codename: 3.7
```
and `rosdep db` yields:
```
OS NAME: alpine
OS VERSION: 3.7
```

Which I think is still not ideal as there's confusion regarding the OS VERSION. But, it does preserve both the true OS version string while providing a simpler codename that can be used to link with releases. So, it a step towards simplifying the inclusion of Alpine in rosdistro, but does break the version string.

If there is no opposition, I would ideally like to have:
`os_detect.py`:
```
OS Name:     alpine
OS Version:  3.7
OS Codename: 3.7
```
`rosdep db`
```
OS NAME: alpine
OS VERSION: 3.7
```

which can be realised by modifying the Alpine class of `os_detect.py` like:
```
class Alpine(FdoDetect):
    """
    Detect Alpine Linux
    """
    def __init__(self):
        super(Alpine, self).__init__('alpine')

    def get_version(self):
        if self.is_os():
            return ".".join(self.release_info["VERSION_ID"].split('.')[:2])
        raise OsNotDetected("called in incorrect OS")

    def get_codename(self):
        if self.is_os():
            return get_version()
```

dirk-thomas:
@tfoote Please review.

MarkHedleyJones:
Fair enough. That makes sense and reflects the point made by @cottsay.
In that case, this pull-request is unnecessary and I will update the one to *rosdep* to reflect this decision.
Thanks for your time.

Issue #176, opened 2019-10-04
OpenSUSE Tumbleweed detection broken (again?)


gavanderhoorn:
I wasn't sure whether to comment on #120 or open a new issue, so we could close this one if re-opening #120 would be preferred.

---

Running `os_detect.py` on Tumbleweed fails again, with the typical error message (identical to what is reported in #120).

Looking at `/etc/os-release`, it would appear that the value for the `ID` key has changed. It was `opensuse` and is now `opensuse-tumbleweed`. As `FdoDetect` performs an equality check of `ID` versus what is registered ([here](https://github.com/ros-infrastructure/rospkg/blob/d5ad8612c47ccb014b6d06de98a4a46ace7ba31a/src/rospkg/os_detect.py#L185) and in this case `opensuse`), this fails, leading to the error.

Reason for opening this issue: would the preferred approach be to register a new key [here](https://github.com/ros-infrastructure/rospkg/blob/d5ad8612c47ccb014b6d06de98a4a46ace7ba31a/src/rospkg/os_detect.py#L717) and then add a new registration [here](https://github.com/ros-infrastructure/rospkg/blob/d5ad8612c47ccb014b6d06de98a4a46ace7ba31a/src/rospkg/os_detect.py#L747), or would something else be better?

I'm hesitant to just submit a PR, as adding keys + registrations for tumbleweed specifically doesn't seem like a scalable approach (ie: need changes in `rospkg` for every new OS release).

gleichdick:
Same here. `CODENAME` definitions are missing in the files, so an exception is thrown in line 265.

### Tumbleweed
Content of `/etc/SUSE-brand`.
```
openSUSE
VERSION = tumbleweed
```
`/etc/os-release`
```
NAME="openSUSE Tumbleweed"
# VERSION="20191011"
ID="opensuse-tumbleweed"
ID_LIKE="opensuse suse"
VERSION_ID="20191011"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:tumbleweed:20191011"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
LOGO="distributor-logo"
```

### Leap 15.0
`/etc/SUSE-brand`:
```
openSUSE
VERSION = 15.0
```

dirk-thomas:
@tfoote Friendly ping.

[...]

PR #179, opened 2019-10-14
Fix openSUSE leap detection

gleichdick:
File /etc/SuSE-release is absent in recent openSUSE releases (leap 15.0, tumbleweed), so ignore it on first try.

Fix for #176


dirk-thomas:
@tfoote Please review.

PR #181, opened 2019-11-03
Fixed os_detect.py for elementary os

svipindev:
Additional quotes removed for elementary os in LsbDetect file.


dirk-thomas:
Thanks for the patch.

PR #182, opened 2019-11-24
Fixes for Python 3.8

jspricke:

PR #186, opened 2019-12-13
only conditionally declare distro dependency in setup.py

### LINKED ISSUE #185: On Ubuntu 16.04 Xenial, python-rospkg now depends on non-exsting python-distro
```
$ docker pull ubuntu:16.04
[...]
$ docker run -it ubuntu:16.04
# echo 'deb http://packages.ros.org/ros/ubuntu xenial main'  > /etc/apt/sources.list.d/ros-latest.list
# apt-key adv --keyserver hkp://ha.pool.sks-keyservers.net:80 --recv-key C1CF6E31E6BADE8868B172B4F42ED6FBAB17C654
[...]
# apt-get -y update
[...]
# apt-get -y install python-rospkg-modules
[...]
The following packages have unmet dependencies:
 python-rospkg-modules : Depends: python-distro but it is not installable
E: Unable to correct problems, you have held broken packages.
```

dirk-thomas:
Fixes #185.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,8 +8,8 @@ from setuptools import setup
 install_requires = ['catkin_pkg', 'PyYAML']
 
 if (
-    sys.version_info[0] > 3 or
-    sys.version_info[0] == 3 and sys.version_info[1] >= 8
+    'SKIP_PYTHON_MODULES' not in os.environ and
+    'SKIP_PYTHON_SCRIPTS' not in os.environ
 ):
     install_requires.append('distro')
 
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,14 @@ import sys
 
 from setuptools import setup
 
+install_requires = ['catkin_pkg', 'PyYAML']
+
+if (
+    sys.version_info[0] > 3 or
+    sys.version_info[0] == 3 and sys.version_info[1] >= 8
+):
+    install_requires.append('distro')
+
 kwargs = {
     'name': 'rospkg',
     # same version as in:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -5,6 +5,14 @@
 
 from setuptools import setup
 
+install_requires = ['catkin_pkg', 'PyYAML']
+
+if (
+    sys.version_info[0] > 3 or
```
nuclearsandwich: Won't this need import sys?

dirk-thomas: Line 4?

nuclearsandwich: My mistake, sys was neither imported nor used in the diff context.

tfoote: This is a follow up to #182 in that we have the conditional for if it's less that python 3.8. To that end should we have the same check here?

dirk-thomas: @tfoote I don't understand the question. The condition checks exactly that: if the Python version is 3.8 or higher the `distro` dependency is needed.

tfoote: sorry I missed the second half under the comment

setup.py
```diff
@@ -5,6 +5,14 @@
...
+    sys.version_info[0] == 3 and sys.version_info[1] >= 8
```
sloretz: Is the version of python used to call `setup.py` guaranteed to be the same as the version of python on the target system? Does that mean during release the `setup.py` invoked multiple times, once for each target platform?

dirk-thomas: I agree with your concern. The first commit probably doesn't work.

Please see 9627024db19a3832be7ab2058e7811bbd9bf72e4 for an alternative approach.

dirk-thomas:
The latest state only declares `distro` in the `setup.py` as a dependency when not building the Debian packages. For Debian packages the explicit `Depends3` in the `stdeb.cfg` file should be sufficient. :crossed_fingers:

PR #188, opened 2019-12-30
Use /bin/sh instead of python 2

sloretz:
This fixes a test failure in `test.test_rospkg_os_detect.test_osx` on systems that don't have python 2 installed by using `sh` instead of `python`.

https://github.com/ros-infrastructure/rospkg/pull/182#issuecomment-566835331

### DIFF
--- a/test/os_detect/osx/sw_vers
+++ b/test/os_detect/osx/sw_vers
@@ -1,3 +1,2 @@
-#!/usr/bin/env python
-import sys
-sys.stdout.write("10.6.5\n")
+#!/bin/sh
+echo "10.6.5"
--- a/test/os_detect/osx/sw_vers_bad
+++ b/test/os_detect/osx/sw_vers_bad
@@ -1,3 +1,2 @@
-#!/usr/bin/env python
-import sys
-sys.stdout.write("10\n")
+#!/bin/sh
+echo "10"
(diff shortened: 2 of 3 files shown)


dirk-thomas:
Instead of using `/bin/sh` in the shebang which makes this test platform specific I would suggest to invoke the `sw_ver` / `sw_vers_bad` with prefixing the command with the Python interpreter used for the test.

sloretz:
> I would suggest to invoke the `sw_ver` / `sw_vers_bad` with prefixing the command with the Python interpreter used for the test.

Do you mean [the call in `OSX.get_version()`](https://github.com/ros-infrastructure/rospkg/blob/1ade2d895195f76f83f1d2b56952ed5315f9e2fd/src/rospkg/os_detect.py#L390) should be prefixed with `sys.executable`? It looks like [`/usr/bin/sw_vers`](https://github.com/ros-infrastructure/rospkg/blob/1ade2d895195f76f83f1d2b56952ed5315f9e2fd/src/rospkg/os_detect.py#L371) on mini3 is a binary executable, so there would need to be a hook to tell the OSX detector this is a test and it should do something different.

> Instead of using `/bin/sh` in the shebang which makes this test platform specific

This test is already relying on the shebang. How about we disable the test on non-POSIX platforms?

dirk-thomas:
> How about we disable the test on non-POSIX platforms?

Sounds good to me.

sloretz:
Skips failing test on non-posix platforms in 6e482d88bdb8b4abfe2dc2ee5af3c0ebbdbedb96

PR #192, opened 2020-03-27
cElementTree has been deprecated since Python 3.3 and removed in Python 3.9

tirkarthi:
Fixes Fedora builds for Python 3.9 : https://bugzilla.redhat.com/show_bug.cgi?id=1817942

### DIFF
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -31,8 +31,13 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 import os
+import sys
 from threading import Lock
-from xml.etree.cElementTree import ElementTree
+
+if sys.version_info >= (3, 3):
+    from xml.etree.ElementTree import ElementTree
+else:
+    from xml.etree.cElementTree import ElementTree
 
 from .common import MANIFEST_FILE, PACKAGE_FILE, ResourceNotFound, STACK_FILE
 from .environment import get_ros_paths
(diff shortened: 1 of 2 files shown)


tfoote:
I confirmed this is failing in 3.9 in #193 could you rebase this on top of that or merge that into this branch so we can get the confirmation that this is then passing CI for 3.9

tirkarthi:
I have added 3.9-dev since 3.9 is not present. The build passes but the deprecated was removed in Python 3.9 alpha 5 so I am not sure if Travis was updated to use this latest version.

src/rospkg/rospack.py
```diff
@@ -31,8 +31,13 @@
...
 
 import os
+import sys
 from threading import Lock
-from xml.etree.cElementTree import ElementTree
+
+if sys.version_info >= (3, 3):
+    from xml.etree.ElementTree import ElementTree
+else:
+    from xml.etree.cElementTree import ElementTree
```
dirk-thomas: Please change the logic to avoid the version condition. Simply try to import `cElementTree` first and if that raises an `ImportError` fallback to the alternative.

tirkarthi: Thanks, made the changes.

dirk-thomas:
Thanks for the patch and for iterating on it.

Since Python 3.9 is still under development (and the key in Travis isn't `3.9` yet but `3.9-dev` yet) I don't think I want to add it to the test matrix. It is good though that the PR shows that with the change CI passes green with the development version of Python 3.9.

dirk-thomas:
That being said I will merge the fix with the change to the Travis test matrix removed (see 438216556c8e883e38994e03805b471405d2e15d).

tirkarthi:
Thank you :)

PR #193, opened 2020-03-28
add 3.9 to test matrix

tfoote:
Add python 3.9 to travis config

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.6"
   - "3.7"
   - "3.8"
+  - "3.9"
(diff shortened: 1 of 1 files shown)


dirk-thomas:
@tfoote Travis CI didn't run for this PR.

tfoote:
It did run but the travis status notification got lost: https://travis-ci.org/github/ros-infrastructure/rospkg/builds/667928694 It failed as expected. This was specifically to show the necessity of #192 it will be replaced by that which is not adding the test matrix. We can close this now that it's being added over there.

PR #196, opened 2020-04-20
Ensure 'distro' dependency is absent for python < 3.8

cottsay:
This approach aligns with the existing approach for `argparse`, while still maintaining the current behavior when `SKIP_PYTHON_{MODULES,SCRIPTS}` is specified.

This is a generalized solution to the change made in #186

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,14 +5,6 @@ import sys
 
 from setuptools import setup
 
-install_requires = ['catkin_pkg', 'PyYAML']
-
-if (
-    'SKIP_PYTHON_MODULES' not in os.environ and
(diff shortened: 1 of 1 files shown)


cottsay:
Travis is failing because we're getting a prerelease of `pyparsing 3.0.0`.

dirk-thomas:
I just tried to simplify the logic a bit in 67bab6e621c4c697bc2087ea3b4682e2bea45ed6. d36a67b33ae3dfe80a9f41d458dab29ae2317fbb is still necessary (even thought the Debian package depends on `python3-distro` in Python 3.8 and higher) since the same Debian package is used for multiple Ubuntu / Debian distros with different Python versions.

PR #199, opened 2020-05-05
Add Pop! OS detector

emersonknapp:
I use Pop! OS linux distro https://pop.system76.com/
It's Ubuntu-based so it works fine for ros dev. 
But rosdep fails because it can't detect the OS. 
Adding a detector here for it.

Signed-off-by: Emerson Knapp <emerson.b.knapp@gmail.com>

### DIFF
--- a/doc/os_detect.rst
+++ b/doc/os_detect.rst
@@ -26,6 +26,7 @@ Currently supported OSes:
 - Gentoo
 - Mint
 - OS X
+- Pop! OS
 - Red Hat Linux
 - Slackware 
 - Ubuntu
@@ -136,6 +137,10 @@ OS name definitions
 .. data:: OS_OSX
 
    Name used for OS X.
+
+.. data:: OS_POP
+
+   Name used for Pop! OS.
 
 .. data:: OS_RHEL
 
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -134,7 +134,7 @@ class OsDetector(object):
 
 class LsbDetect(OsDetector):
     """
-    Generic detector for Debian, Ubuntu, and Mint
+    Generic detector for Debian, Ubuntu, Mint, and Pop! OS
     """
     def __init__(self, lsb_name, get_version_fn=None):
         self.lsb_name = lsb_name
@@ -731,6 +731,7 @@ OS_OPENSUSE13 = 'opensuse'
 OS_TIZEN = 'tizen'
 OS_SAILFISHOS = 'sailfishos'
 OS_OSX = 'osx'
+OS_POP = 'pop'
 OS_QNX = 'qnx'
 OS_RHEL = 'rhel'
 OS_SLACKWARE = 'slackware'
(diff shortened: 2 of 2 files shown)


emersonknapp:
Question: with this change, rosdep doesn't choose the "ubuntu" option for resolving dependencies, is that something that would happen here? I notice there is a Mint matcher, but don't see any Mint in the rosdistro rules - how does that work? 

Rosdep succeeds in `rosdep update` with this change, but then it says 

```
Unsupported OS: pop
Supported OSes are [alpine, arch, cygwin, debian, ubuntu, gentoo, nixos, openembedded, opensuse, osx, fedora, rhel, slackware, freebsd]
```

when I try to `rosdep resolve` a package

mikaelarguedas:
Love it when I face an issue an I find a PR already opened for it :heart: 

> But rosdep resolve now [...]

I opened https://github.com/ros-infrastructure/rosdep/pull/757 that should solve that problem and give the following output:

```
$ rosdep resolve python3-twisted
WARNING: ROS_PYTHON_VERSION is unset. Defaulting to 3
rosdep detected OS: [pop] aliasing it to: [ubuntu]
#apt
python3-twisted
``` 

> rosdep install --simulate --os=ubuntu:focal python3-twisted  
ERROR: Rosdep cannot find all required resources to answer your query
Missing resource python3-twisted

This one looks more like a misuse of `rosdep install`.
Some threads on ROS answers give some explaination about this e.g. https://answers.ros.org/question/348230/how-do-i-use-rosdep-install-to-install-individual-dependencies/

emersonknapp:
Oh - perfect! I didn't know about the aliasing logic in rosdep - I was just trying to get install to work, I can see that the `--os` override is not the right way to go.

mikaelarguedas:
> Oh - perfect! I didn't know about the aliasing logic in rosdep - I was just trying to get install to work, I can see that the --os override is not the right way to go.

OS override is a totally valid way of doing it. The aliasing is just syntactic sugar on top to avoid having to pass the `--os` flag.

The following should work even with no alias defined
```
mkdir -p /tmp/ws && cd /tmp/ws
git clone https://github.com/eclipse-cyclonedds/cyclonedds src/cyclonedds
rosdep install -s --from-paths src --ignore-src --os=ubuntu:focal
#[apt] Installation commands:
  sudo -H apt-get install libcunit1-dev
```

The confusing part is that `rosdep install` never takes rosdep keys as arguments (like `python3-twisted`) but only operates on ROS packages on your path or paths to ros packages.

emersonknapp:
Oh! Yeah ok I knew this, wasn't thinking about it - I just pattern-matched from `resolve` to `install`.

emersonknapp:
I can attest that it's working on my pop 20.04. I think it was not relevant in pre-20 releases because it reported "Ubuntu" for that field of lsb_release.

I also can't find a docker image of pop. Makes sense since it's mostly just the desktop environment, but yeah that makes it hard to test. Maybe just second point of validation from @mikaelarguedas is good?

mm318:
Hi @mikaelarguedas, did you get a chance to test this fix on your Pop! OS environment?

mikaelarguedas:
> Hi @mikaelarguedas, did you get a chance to test this fix on your Pop! OS environment?

Yes, I used it happily on May 6th along https://github.com/ros-infrastructure/rosdep/pull/757. Didnt have much need to run rosdep since then

mikaelarguedas:
Friendly request for a new release of rospkg including this PR to unlock https://github.com/ros-infrastructure/rosdep/pull/757

dirk-thomas:
I was waiting for another pending PR but since it didn't get any attention yet I just :ship:-ed `1.2.7`.

mikaelarguedas:
Thanks! :ship:

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-07-16/15468/1

PR #200, opened 2020-05-26
Add EulerOS to os_detect.py

yanqd0:
EulerOS: <https://developer.huaweicloud.com/ict/en/site-euleros/euleros>


dirk-thomas:
@tfoote Friendly ping.

PR #204, opened 2020-06-23
Add Python2-Depends-Name option for stdeb 0.9.0.

nuclearsandwich:
This option is required to make releases from Ubuntu Focal and other
distributions where the Python 2 package name is python2 rather than
python.

This option requires the use of stdeb 0.9.0 or later. https://github.com/astraw/stdeb/pull/156

https://github.com/ros-infrastructure/catkin_pkg/pull/288 is a parallel change for catkin-pkg.

Python 2 packages from the recent 1.2.7 release are currently uninstallable as a result of this option being unset.

### DIFF
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -12,6 +12,7 @@ Conflicts: python3-rospkg
 Conflicts3: python-rospkg
 Suite: xenial yakkety zesty artful bionic cosmic disco eoan jessie stretch buster
 Suite3: xenial yakkety zesty artful bionic cosmic disco eoan focal jessie stretch buster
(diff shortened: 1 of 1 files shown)


ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/build-ci-failures-with-ros-testing-repository-enabled/15006/4

nuclearsandwich:
One mistake, the option was introduced in 0.9.1.

dirk-thomas:
:+1: afterwards. Thanks for making a patch release.

Issue #202, opened 2020-06-07
Test failing on ARM7 Alpine Linux


russkel:
Hi, I am trying to get this package compiling on Alpine Linux and it seems to be failing on one test on an arm7 machine https://gitlab.alpinelinux.org/russkel/aports/-/jobs/137312

It seems on this machine the arch is `armv7`. Should this be added to the list?
```
 ======================================================================
FAIL: test.test_rospkg_os_detect.test_tripwire_uname_get_machine
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/builds/russkel/aports/testing/py3-rospkg/src/rospkg-1.2.6/test/test_rospkg_os_detect.py", line 413, in test_tripwire_uname_get_machine
    assert retval in [None, 'aarch64', 'armv7l', 'i386', 'i686', 'ppc', 'ppc64', 'ppc64le', 's390', 's390x', 'x86_64']
AssertionError
----------------------------------------------------------------------
```

https://github.com/ros-infrastructure/rospkg/blob/9c1dc915f27a391198dc61cfe4025dc39c934dfa/test/test_rospkg_os_detect.py#L413

dirk-thomas:
@tfoote Can you please take a look at this.

tfoote:
It looks like we need to extend that list. Can you validate that adding the 'armv7' to that list lets it pass?

russkel:
@tfoote turns out it was `armv8l` which is what the `armv7` builder is reporting as. I have added and tested this in #206

russkel:
Solved!

PR #206, opened 2020-07-10
tests: add armv8l arch to tripwire_uname test

russkel:
Should solve #202 for Alpine Linux support. Tested on Alpine Linux armv7 builder.


russkel:
It's actually `armv7`.

russkel:
Thanks gents.

PR #208, opened 2020-09-09
Add support for macOS Big Sur

rotu:


cottsay:
Two thoughts:
1. None of the hundreds of `osx` rosdep keys are going to work on a platform that detects the OS as `macos`, regardless of whether the same `OSDetector` class is used. You should use an alias, as is done for derivatives of Ubuntu and RHEL.
2. I'm not thrilled to see a mixture of refactor changes mixed in with the change you're proposing. If you believe moving stuff into that class provides value, I think you should open a separate PR, and we'll discuss whether it is worth touching the lines and complicating a `git blame` process.

cottsay:
> None of the hundreds of `osx` rosdep keys are going to work on a platform that detects the OS as `macos`, regardless of whether the same `OSDetector` class is used. You should use an alias, as is done for derivatives of Ubuntu and RHEL

Nix that, the aliasing happens in `rosdep`, not here. Mea culpa.

rotu:
Everything past Sierra is macOS, and Big Sur is not even macOS 10. Feel free to propose a better fix or only pull some of these changes.

rotu:
extracted to #209 and #210

PR #212, opened 2020-11-12
Use GitHub Actions for rospkg CI.

nuclearsandwich:
With the exception of 3.4 which is not supported by the setup-python
action so we leave that one set up in Travis for the (likely brief)
remainder of python 3.4 support.

This is mostly a port of the travis handling. I did change `nosetests` to `python -m nose`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,7 @@ name: rospkg-ci
 
 on:
   push:
-    branches: ['master']
+    branches: [master]
   pull_request:
     branches: ['*']
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -45,7 +45,7 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip setuptools
-          python -m pip install nose coverage
+          python -m pip install nose coverage mock
           python setup.py build develop
       - name: Run tests
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,61 @@
...
+            python -m pip install pyparsing==2.4.7
+          fi
+          python -m pip install --upgrade pip setuptools
+          python -m pip install nose coverage mock
+          python setup.py build develop
+      - name: Run tests
+        run: |
+          python -m nose --with-coverage --cover-package=rospkg --with-xunit test
+
+
```
cottsay, line 61: Not a huge fan of excess trailing whitespace...

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,61 @@
+name: rospkg-ci
+
+on:
+  push:
+    branches: ['master']
```
cottsay, line 5: You didn't use quotes in the other PR. It would be nice to be consistent - I don't care which way.

cottsay (approved):
Looks fine, a couple of notes, but no blockers.

.github/workflows/ci.yaml
nuclearsandwich: Hah. What comes of making edits while being social at ROS World. Dropped quotes.
```suggestion
    branches: [master]
```

.github/workflows/ci.yaml
nuclearsandwich: Thanks for spotting.

.github/workflows/ci.yaml
nuclearsandwich: Thanks for spotting.

PR #213, opened 2020-12-01
os_detect: support macOS Big Sur

okalachev:
`rosdep update` is now failing on macOS Big Sur.

This commit makes it not to fail and detect it correctly.

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -343,25 +343,27 @@ class Rhel(Fedora):
 
 # Source: https://en.wikipedia.org/wiki/MacOS#Versions
 _osx_codename_map = {
-    4: 'tiger',
-    5: 'leopard',
-    6: 'snow',
-    7: 'lion',
-    8: 'mountain lion',
-    9: 'mavericks',
-    10: 'yosemite',
-    11: 'el capitan',
-    12: 'sierra',
-    13: 'high sierra',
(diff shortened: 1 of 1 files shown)


bjajoh:
@dirk-thomas @tfoote Please merge if possible, I'm also highly interested in this fix. :)

dirk-thomas:
@bjajoh I am not the maintainer of this repository anymore. Hopefully the new maintainer will provide you with feedback soon (unfortunately I don't recall who that is).

rtwomey:
I believe this will cover Big Sur through the beta, but [the current latest version is 11.1](https://en.wikipedia.org/wiki/MacOS_Big_Sur). As of this writing, running `sw_vers -productVersion` unfortunately returns `11.1`. I guess that means "Big Sur" is the codename for both 11.0 and 11.1....

smnogar:
> I believe this will cover Big Sur through the beta, but [the current latest version is 11.1](https://en.wikipedia.org/wiki/MacOS_Big_Sur). As of this writing, running `sw_vers -productVersion` unfortunately returns `11.1`. I guess that means "Big Sur" is the codename for both 11.0 and 11.1....

I just pushed [a new PR](https://github.com/ros-infrastructure/rospkg/pull/216) that should address this issue.

tfoote:
Closing in favor of #216

PR #215, opened 2020-12-17
Fix CentOS 8.3 detection

cottsay:
It seems that CentOS 8.3 dropped the codename part of the `redhat-release` file entirely. Even before this change, the codename was essentially a placeholder - rosdep is configured to use the version number (major/minor only). Since there isn't anything in parenthesis in the file, just return the last "word" on the line, which should be the version number.

Also mimic the RHEL tests for CentOS to verify this change, and add a RHEL 8 release file to the existing RHEL tests.


nuclearsandwich (approved):
Less than thrilled that this broke in the middle of a release but the updated logic is more robust and has tests to back it up.

PR #216, opened 2021-01-04
fix for detecting big sur (and newer) versions of macOS

smnogar:
A slight tweak to okalachev's PR that should work for Big Sur 11.X and future macOS releases.

PR #217, opened 2021-02-10
Add compatibility for MacOS Big Sur (11.0.X) -> Update os_detect.py

Piit3r:
Add compatibility for MacOS Big sur (11.0.X)

Avoid the error :
> rospkg.os_detect.OsNotDetected: unrecognized version: 11.0

When running
`rosdep update`

I don't know if there will be other errors on Big Sur


christophebedard:
Just so you know, there's a PR open for this already: https://github.com/ros-infrastructure/rospkg/pull/216

Also, it seems that Big Sur is 11.x and not 11.0.x, so your solution might not work.

PR #218, opened 2021-03-01
Add "opensuse-leap" support to os_detect.py

neotinker:
The correct ID value for openSUSE Leap is "opensuse-leap"

localhost:~ # grep -E "^ID=|^VERSION_ID=" /etc/os-release
ID="opensuse-leap"
VERSION_ID="15.2"
localhost:~ #


cottsay (approved):
Tested using the Docker container `opensuse/leap`:
```
# python3 -m rospkg.os_detect
OS Name:     opensuse
OS Version:  15.2
OS Codename: 15.2
```

nuclearsandwich (approved):

PR #220, opened 2021-04-14
ManifestManager.list() should return a list

ivanpauno:
In python3 this is returning a "dict_keys", that for example broke code like this:

```python3
self.allnames = sorted(list(set(rospack.list() + rosstack.list()))) 
```

because the `+` operator isn't supported by `dict_keys`.

See [this](https://github.com/ros-visualization/rqt_dep/issues/15#issuecomment-819361754) and [this](https://github.com/ros-visualization/rqt_dep/issues/15#issuecomment-819527264) comments.
Downstream packages can also fix this directly, but it would be better if the `list()` method actually returned a python `list`.

### DIFF
--- a/src/rospkg/rospack.py
+++ b/src/rospkg/rospack.py
@@ -194,7 +194,7 @@ class ManifestManager(object):
         :returns: complete list of package names in ROS environment, ``[str]``
         """
         self._update_location_cache()
-        return self._location_cache.keys()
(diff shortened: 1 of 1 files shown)


wxmerkt:
@cottsay Friendly ping - any thoughts on this PR? Tools like `rqt_dep` are broken in Noetic and having this fix incorporated would go a long way in getting them back operational :-).

PR #221, opened 2021-04-27
Add option to show all ROS package names and their versions

708yamaguchi:
I add option to show all ROS package names and their versions like below.
I think this option is useful for printing an overview of the ROS environment.
```
$ rosversion --all
...
rospack: 2.5.6
rospy: 1.14.10
rviz: 1.13.15
...
```

Without this option, we need to do like below and this takes a little long time.
```
$ for i in $(rospack list-names); do echo "$i: $(rosversion $i)"; done
```

### DIFF
--- a/src/rospkg/rosversion.py
+++ b/src/rospkg/rosversion.py
@@ -69,6 +69,29 @@ def print_without_newline(argtext):
     print(argtext, end='')
 
 
+def get_version_from_package_name(package):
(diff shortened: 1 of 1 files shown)


cottsay (approved):
This seems like a reasonable change to me. CI is broken in this repository right now, but once we get that fixed and get some green builds on this PR, I think it will be good to go.

Thanks!

PR #224, opened 2021-05-03
Add Conda OS detection.

Tobias-Fischer:
Hi there,
We are currently adding `rosdep` functionality to the [RoboStack](https://github.com/RoboStack/ros-noetic) project which builds ROS binaries with Conda. The first step towards this aim is adding an `OsDetector` here. Locally, we already have a working version of `rosdep` that makes use of the new RoboStack OsDetector, and would like to merge as many changes as possible upstream. See https://github.com/RoboStack/ros-noetic/issues/41 for more information

/cc @wolfv @traversaro

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -242,12 +242,14 @@ class OpenEmbedded(OsDetector):
         raise OsNotDetected('called in incorrect OS')
 
 
-class RoboStack(OsDetector):
+class Conda(OsDetector):
     """
-    Detect RoboStack.
+    Detect Conda.
     """
     def is_os(self):
-        return "ROS_OS_OVERRIDE" in os.environ and os.environ["ROS_OS_OVERRIDE"].startswith("robostack")
+        return "ROS_OS_OVERRIDE" in os.environ and
+            (os.environ["ROS_OS_OVERRIDE"].lower().startswith("robostack") or
+             os.environ["ROS_OS_OVERRIDE"].lower().startswith("conda"))
 
     def get_version(self):
         if self.is_os():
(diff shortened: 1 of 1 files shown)


nuclearsandwich:
As far as I know this is the first time that OS detection would be used to switch to the installer on top of another platform. My immediate concern is that this isn't composable. That is, you can't use a mix of Ubuntu and RoboStack dependencies. While doing so is probably a bad idea, we should think about how other cross-platform environments will look at robostack as a precedent.

What is the advantage of RoboStack being an OS in rospkg/rosdep rather than an installer available on multiple platforms?

traversaro:
> As far as I know this is the first time that OS detection would be used to switch to the installer on top of another platform.

Just to give you the full picture, this modification was inspired by https://github.com/ros-infrastructure/rospkg/pull/171 done by @lopsided98 . If I understood correctly, in that case, `Nix` could be used either as the package manager of the `NixOs` operating system, as a "user" package manager on any Linux distribution, a bit similar to how [`conda`](https://docs.conda.io/en/latest/) works.

Tobias-Fischer:
Yes - fully agree with what @traversaro said. We already follow precedent - that of `NixOs`, which seems like it is very similar to our `conda` use case; in particular, `NixOs` is a multi-platform package manager as well.

Tobias-Fischer:
Change RoboStack to Conda, and change OS_OVERRIDE to check whether we are in a `$CONDA_PREFIX` and `rosdep` is in `sys.prefix`

nuclearsandwich (approved):
After reviewing the nixOS support in rospkg (I hadn't done so closely before) I think this makes sense as well. Thanks for iterating!

cottsay (approved):

PR #229, opened 2021-10-26
Update os_detect.py to add macOS Monterey

benhjames:
Monterey was released yesterday, so this adds it to the macOS codename map!

Closes #232


christophebedard:
@cottsay does this PR need another review, or can it be merged?

PR #231, opened 2021-10-29
Remove legacy rosbuild files

cottsay:
This package is distributed as a standalone Python package in all versions of ROS, and has been for many releases. This manifest and CMakeLists.txt are from the transitional days around when catkin first emerged, when it appears that rospkg was distributed as a ROS package instead of a standalone python package.

The presence of a 'dry' manifest is confusing colcon and serves no useful purpose anymore.


cottsay:
Oops, looks like the tests need to be updated. I'll make a fix.

nuclearsandwich (approved):

PR #232, opened 2021-10-30
fix: Add macOS monterey to os_detect

PymZoR:
Similar issue https://github.com/ros-infrastructure/rosdep/issues/615
`rosdep update` is failing on macOS monterey


christophebedard:
#229 should take care of this

PR #234, opened 2021-11-22
fix test for riscv machines (64 and 32 bits)

alexfanqi:
Hi, there.

I am trying to build ROS on Gentoo for riscv natively, specically on a Sifive Unmatch board. As riscv is an emerging arch, I guess soon others may also want to build ros packages natively on riscv. It would be a good addition to make this test pass.


nuclearsandwich:
I didn't see this run by the first time but I'm adding it to the release notes and heck yeah RISC V!

PR #238, opened 2021-12-07
Update debian codename mapping

cottsay:
A few issues here:
1. With py3.8, we're using 'distro' instead of 'platform', which gives us only the major version number. Using str.split handles both cases.
2. Add bookworm and trixie
3. Looks like 'sid' doesn't have a number, but that's okay

This should only result in a change in behavior for Debian versions newer than Buster. I tested it on Buster, Bullseye and Sid.

### DIFF
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -169,16 +169,17 @@ class Debian(LsbDetect):
 
     def get_codename(self):
         if self.is_os():
-            v = self.get_version()
-            if v.startswith('7.'):
-                return 'wheezy'
-            if v.startswith('8.'):
-                return 'jessie'
-            if v.startswith('9.'):
-                return 'stretch'
-            if v.startswith('10.'):
-                return 'buster'
-            return ''
+            v = self.get_version().split('.', 1)[0]
+            return {
+                '7': 'wheezy',
+                '8': 'jessie',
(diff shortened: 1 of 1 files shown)


nuclearsandwich (approved):

nuclearsandwich:
Actually, does it make sense to switch Debian to use the Freedesktop.org /etc/os-release detector? It might still need some special handling for Sid as here's what I see on Bullseye vs Sid for `/etc/os-release`:

```
lxc exec bullseye /bin/cat /etc/os-release                                                                                  (base)
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

```
❯ lxc exec sid /bin/cat /etc/os-release                                                                                           (base)
PRETTY_NAME="Debian GNU/Linux bookworm/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

PR #243, opened 2022-02-18
Require distro 1.6.x with Python 2

cottsay:
Support for Python 2 was [officially dropped in 1.7.0](https://github.com/python-distro/distro/blob/master/CHANGELOG.md#170-20220215).


cottsay:
Closing in favor of #245

PR #244, opened 2022-02-18
Add OS_RASPBIAN

cottsay:
We should set up an alias in rosdep as soon as possible so that when we're ready to switch raspbian to detecting separately from Debian, we don't break rosdep.

The steps to getting where we want to be:
1. Merge this PR to make `OS_RASPBIAN` available and release `rospkg`
2. Update `rosdep` to alias `OS_RASPBIAN` to `OS_DEBIAN` and release `rosdep`
3. Ensure the new `rosdep` is widely available
4. Update `rospkg` to detect Raspbian as `OS_RASPBIAN`

We may want to make `rospkg` conflict with `rosdep` versions which don't have the `OS_RASPBIAN` alias. We don't have the ability to do that in pip, but we could at least do it in the debs and RPM packages. It might save someone a headache at some point.


tfoote (approved):
This looks good. The risk of compatibility is if the user updates rospkg detection but doesn't update rosdep to do the aliasing. To that end a conflict with rosdep less than aliasing does guarantee accuracy, it doesn't get every case, but will get the majority and hopefully people running from pip are generally updating everything or pinning everything.

nuclearsandwich (approved):

PR #245, opened 2022-02-18
Make dependency on distro specific to py3.8+

cottsay:
The code only uses the `distro` module if using Python 3.8 and newer. The dependency should reflect that to avoid unnecessarily installing the dependency.


nuclearsandwich (approved):
This just bit me in CI since 2.7 tries to install distro and isn't compatible with type syntax.

PR #246, opened 2022-02-18
Enable Python 3.10 tests, bump actions/setup-python

cottsay:
This also aligns the workflow better with catkin_pkg.


cottsay:
Evidently this will depend on #247 because it looks like `nose` is broken with Python 3.10.

nuclearsandwich (approved):

PR #248, opened 2022-02-24
Avoid use of deprecated function in distro module.

### LINKED ISSUE #241: Ubuntu detection broken with 'distro>=1.7.0'
```
$ python3 os_detect.py
OS Name:     ubuntu
OS Version:  20.04
OS Codename: focal
$ pip3 install -q --user -U 'distro>=1.7.0'
$ python3 os_detect.py
OS Name:     ubuntu
OS Version:  20.04
OS Codename: Focal Fossa
```

First reported as ros-infrastructure/rosdep#779

nuclearsandwich:
Recent versions of the distro module have deprecated the `distro.linux_distribution()` which is a compatibility shim for the
`platform.linux_distribution()` function removed in python 3.8.

All versions of the distro library going back to at least 1.0.1[1] packaged in Ubuntu Bionic[2] provide the `id`, `version`, and `codename` methods which are recommended instead.

`distro.id()` is used instead of `distro.name()` corresponding with the full_distribution_name=0 argument to `linux_distribution`.

When updating the tests to support this change I noticed that the mocks were being added in such a way that the `dist` conditional branch would never actually be tested. I didn't actually look at how far back we'd have to
go to find a platform module which did not support `linux_distribution` but it made sense to at least make testing that branch possible.

We'll have to take care that the `codename` member of the tuple continues to match what our scripts expect downstream since the tests mock the distro return values we don't have full coverage of the actual results on contemporary platforms.

Closes #241

[1]: https://github.com/python-distro/distro/blob/v1.0.1/
[2]: https://packages.ubuntu.com/bionic/python3-distro

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,7 +10,7 @@ if (
     'SKIP_PYTHON_MODULES' not in os.environ and
     'SKIP_PYTHON_SCRIPTS' not in os.environ
 ):
-    install_requires.append("distro; python_version >= '3.8'")
+    install_requires.append("distro >= 1.4.0; python_version >= '3.8'")
 
 kwargs = {
     'name': 'rospkg',
--- a/src/rospkg/os_detect.py
+++ b/src/rospkg/os_detect.py
@@ -138,7 +138,9 @@ class LsbDetect(OsDetector):
     """
     def __init__(self, lsb_name, get_version_fn=None):
         self.lsb_name = lsb_name
-        if hasattr(distro, "linux_distribution"):
+        if distro.__name__ == "distro":
+            self.lsb_info = (distro.id(), distro.version(), distro.codename())
+        elif hasattr(distro, "linux_distribution"):
             self.lsb_info = distro.linux_distribution(full_distribution_name=0)
         elif hasattr(distro, "dist"):
             self.lsb_info = distro.dist()
--- a/stdeb.cfg
+++ b/stdeb.cfg
@@ -18,7 +18,7 @@ Setup-Env-Vars: SKIP_PYTHON_MODULES=1
 
 [rospkg_modules]
 Depends: python-catkin-pkg-modules, python-yaml, lsb-release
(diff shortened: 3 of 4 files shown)


nuclearsandwich:
Temporarily cherry-picking https://github.com/ros-infrastructure/rospkg/pull/245 to get CI results back.

nuclearsandwich:
I had to do some work in https://github.com/ros-infrastructure/rospkg/issues/241#issuecomment-1050298082 to convince myself that aside from getting us away from a deprecated function this also fixes that issue. At least in all known situations it resolves that issue as well.

cottsay (approved):
Thanks for spearheading the development and testing on this one.

nuclearsandwich:
At @cottsay's prompting I went a little further back and found that there are some cases in older versions of distro where codename _isn't_ what we want.

> However, across all tested versions of the distro module the `distro.codename()` function returns what we want and so the patch in #248 should resolve our issues on all currently supported platforms.

I ran an extra test using the packaged version of `distro` in Ubuntu Bionic, our oldest supported distro and the `distro.codename()` returns `"Bionic Beaver"` which means that I need to update that to say that across all _recent_ versions of distro...

I'm going to update this PR to require distro>=1.4.0 which is when codename first started returning the symbolic name rather than the full nickname.

tfoote (approved):

setup.py
```diff
@@ -10,7 +10,7 @@
     'SKIP_PYTHON_MODULES' not in os.environ and
     'SKIP_PYTHON_SCRIPTS' not in os.environ
 ):
-    install_requires.append("distro; python_version >= '3.8'")
+    install_requires.append("distro >= 1.4.0; python_version >= '3.8'")
```
nuclearsandwich, line 13: @cottsay I've never combined a package version constraint and a python version constraint. This parsed and worked for me locally on Arch Linux but I'd like a second look in case I'm missing a subtlety or style thing.

cottsay: A first for me, too. Seems right, though :+1:

cottsay (approved):

PR #249, opened 2022-02-24
Fix syntax error in Conda detector.

nuclearsandwich:
This slipped in from #224 since it was contributed while CI had other issues going on.

I think there are several different ways to structure this but this was the lightest touch. I'm not too particular if there are other style preferences.


cottsay (approved):
Works for me, thanks for following up.

PR #251, opened 2022-03-07
Declare test dependencies in [test] extra

cottsay:
Also introduce a constraints file to handle Python 2 in CI.


nuclearsandwich:
> Also introduce a constraints file to handle Python 2 in CI.

As I understand it, these constraints are for indirect dependencies which is why they aren't expressed directly in setup.py.
Does it make sense to place the file at the package root so that it's more readily discoverable for others building from source (especially in python2 environments)?

cottsay:
> Does it make sense to place the file at the package root...?

Eh, I could go either way. We could probably do that and then just drop it when we drop Python 2 support.

I'm 50/50 on this, so you'll have to be the deciding vote.

nuclearsandwich:
> We could probably do that and then just drop it when we drop Python 2 support.

Yeah that seems reasonable to me.

> I'm 50/50 on this, so you'll have to be the deciding vote.

My overarching preference is for CI actions to "just run repository scripts" and thus the only contents stashed into CI subdirectories ought to be CI-only so let's move the constraints.txt file to the repo root.

nuclearsandwich:
> so you'll have to be the deciding vote.

by the by I very much appreciate making this explicit since my default tiebreaker is "implementing maintainers choice"

.github/workflows/ci.yaml
```diff
@@ -27,12 +27,7 @@ jobs:
           python-version: ${{matrix.python}}
       - name: Install dependencies
         run: |
-          if test ${{matrix.python}} = 2.7; then
-            python -m pip install pyparsing==2.4.7 'PyYAML<6.0'
-          fi
-          python -m pip install --upgrade pip setuptools
-          python -m pip install mock pytest pytest-cov
-          python setup.py build develop
+          python -m pip install -U -e .[test] pytest-cov -c .github/workflows/constraints.txt
```
nuclearsandwich, line 30: Actually out of curiosity why isn't `pytest-cov` a test dependency?

cottsay: It isn't *actually* needed to run the tests locally, but is something we're seem to be interested in seeing in GitHub Actions.

cottsay:
> ...let's move the constraints.txt file to the repo root.

Done in 50636cf.

.github/workflows/ci.yaml
cottsay: To further clarify, it isn't needed to run `pytest` on this repository unless you manually add `--cov` to the invocation. So I would deem it the caller's responsibility to install it if they'd like to use it.

nuclearsandwich: Part of me wants to advocate for an additional `extras_require.coverage` section but I'll hold my peace until the exact moment that there is more than a single coverage-specific dependency.

nuclearsandwich (approved):
Thanks for addressing review suggestions and answering questions!

PR #256, opened 2023-01-08
Add ventura in the _osx_codename_map

metanav:
This PR will add the latest MacOS version/name in the map.


omk1kor:
Tested on MacOs 13.2, os_detect now reports correct output.

% python3 -m rospkg.os_detect                      
OS Name:     osx
OS Version:  13.2
OS Codename: ventura

cottsay (approved):

wjwwood:
Ran into this today. I suppose we still need a release (1.4.0 is the latest and doesn't contain this)?

PR #257, opened 2023-02-21
Use FdoDetect to identify Raspbian systems

cottsay:
Raspbian Buster was close enough to vanilla Debian Buster that the implementation of the Debian detector in os_detect identified it as such. Things changed in Raspbian Bullsye, and it became necessary to start identifying Raspbian as a discrete platform. This commit makes that change.

For Raspbian Bullseye, this is not a breaking change because the previous behavior was to fail to detect. On Raspbian Buster, this change will break rosdep versions prior to 0.22.0, where the alias was added.

Another note is that the FdoDetect implementation doesn't output the minor version number, but I'm not aware of any downstream code using the version number of Debian platforms at all.

Raspbian Buster before this change:

<details>

```
pi@raspberrypi:~ $ python3 -m rospkg.os_detect
OS Name:     debian
OS Version:  10.13
OS Codename: buster
```

</details>

Raspbian Buster after this change:

<details>

```
pi@raspberrypi:~ $ python3 -m rospkg.os_detect
OS Name:     raspbian
OS Version:  10
OS Codename: buster

pi@raspberrypi:~ $ python3 -m rosdep2 update -q
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
```

</details>

Raspbian Bullseye after this change:

<details>

```
pi@raspberrypi:~ $ python3 -m rospkg.os_detect
OS Name:     raspbian
OS Version:  11
OS Codename: bullseye

pi@raspberrypi:~ $ rosdep update -q
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
rosdep detected OS: [raspbian] aliasing it to: [debian]
```

</details>


nuclearsandwich (approved):

nuclearsandwich:
> Another note is that the FdoDetect implementation doesn't output the minor version number, but I'm not aware of any downstream code using the version number of Debian platforms at all.

To the best of my knowledge nothing does need this. In nearly all cases these versions are interoperable and mostly indicate a minimum ceiling for updates. It is possible for packages to work in one "version" and not another due to package removals or additions as a result of critical bugs. However, a more accurate detection of that problem would be with package dependency expressions themselves and so the minor version for a debian release ought not be needed by rosdep.

PR #260, opened 2023-04-24
Add "openeuler" support to os_detect.py

Z572:


src/rospkg/os_detect.py
```diff
@@ -265,6 +265,25 @@ def get_codename(self):
...
+    Detect OpenEuler OS.
+    """
+    def __init__(self, release_file="/etc/openEuler-release"):
+        self._release_info = read_os_release()
+        self._release_file = release_file
+    def is_os(self):
+        return os.path.exists(self._release_file)
+    def get_version(self):
+        if self.is_os():
+            return self._release_info["VERSION_ID"]
```
cottsay, line 279: Would you mind updating this to prevent the `KeyError` from raising if the release file doesn't contain `VERSION_ID`?

I don't have any strong feelings about how to handle the lack of VERSION_ID, but a `KeyError` isn't very actionable. I think either returning an empty string or raising `OsNotDetected` would probably be appropriate.

PR #262, opened 2023-12-30
add macOS Sonoma to _osx_codename_map

kylegibson15:


christophebedard:
Thanks for the PR! Could you run `python3 -m rospkg.os_detect` and provide the output just to confirm that it does indeed work?

kylegibson15:
```
OS Name:          osx
OS Version:       14.2.1
OS Codename: sonoma
```

cottsay (approved):

PR #264, opened 2024-02-10
Update CI testing matrix

cottsay:
* Drop Python 2.7
* Switch Python 3.6 to Focal
* Add Python 3.11, 3.12
* ~~Add Windows~~

PR #265, opened 2024-02-10
Add detection logic for Fedora Asahi Remix

cottsay:
~~Requires #264~~


tfoote:
Is there a docker image that this can be tested or validated against?

tfoote (approved):
Overall this looks fine. Without an image to test against I can't validate this will work. But it shouldn't break anything.

cottsay:
I don't have actual hardware with which to test this, but I derived the ID from the `os-release` source in the RPM spec file: https://copr-dist-git.fedorainfracloud.org/cgit/@asahi/fedora-remix-branding/fedora-asahi-remix-release.git/tree/fedora-asahi-remix-release.spec#n1377

PR #266, opened 2024-03-01
Update debian codename mapping

cottsay:
* Debian 14 will be called 'Forky'
* Move 'Rodete' to 'Trixie' now that 'Bookworm' is stable


tfoote (approved):
That's going to be a fun naming collision when searching for help on forums.

+1 for keeping rodete moving forward before things diverge.

